PR #14, opened 2021-11-12
Revised API Specification

rj-adams:
Change endpoints to use query parameters where appropriate
Add specific endpoints for filtered get visits requests for ease of use
Add tags

@Lucy. Once we agree this API specification , it is _exactly_ the API specification the back end code should create.

This will mean that the front end can be developed against this specification in advance of the back end existing and in confidence that the api exposed by back end services will look exactly like that specified..

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
 # visit-scheduler
 
-API to provide visit scheduling for the Book a prison visit services
+Back end microservice providing visit scheduling for Book a Prison Visit Services.
+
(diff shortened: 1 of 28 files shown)


psoleckimoj:
There's no file changes to merge?

I agree this is a working document but we can't nail down an API exactly before the front-end has even started. It'll be an agile, iterative process creating the API as we go on. We can use this as a basis to start from.

We're going to be working closely together to develop the backend and frontend so it's not going to cause us any problems. We'll no doubt decide we need a field in a particular format, or another filter or another end point we've not thought of.

rj-adams:
incorrect submission

PR #19, opened 2021-11-24
vb-315 specification change for visit filtering

lbennett-moj:


visit-scheduler-api-specification.yaml
```diff
@@ -13,7 +13,7 @@ servers:
           - preprod   #Pre-production
           - Prod      #Production
 paths:
-  /bookable-visit-sessions-by-prison/{prisonId}:
+  /visit-sessions/prison/{prisonId}:
     get:
```
lbennett-moj, line 17: this may change to a query param style but just updating the resource for now

rj-adams:
Yes I think we need the param style change also - as we've agreed to go with filters.

I would expect this endpoint to now be something similar to:

`/visit-sessions?prison={prisonID}`

psoleckimoj (approved):
LGTM!

PR #22, opened 2021-12-01
refactor "establishment" to "prison" in API specification

rj-adams:


lbennett-moj (approved):

visit-scheduler-api-specification.yaml
```diff
@@ -429,13 +429,13 @@ components:
...
+      description: prison Description
       properties:
-        establishmentId:
+        prisonId:
           type: string
-          description: Unique identifier for this establishment
-        establishmentName:
+          description: Unique identifier for this prison
+        prisonName:
           type: string
```
lbennett-moj, line 440: Discussion point.  The front end service will be looking up descriptions and reference data - should prison name lookup be done there too?

lbennett-moj (approved):

visit-scheduler-api-specification.yaml
rj-adams: Looking at the response of other API endpoints, they seem to always return both the code and the description - so I followed suit - it doesn't seem worth the effort of the extra call to get the name.

lbennett-moj:
I would recommend only storing ids or business keys  in our database and
looking up description data mastered in nomis - prison-api endpoints do
tend to return them because they are looking them up directly from nomis.
Can discuss in the friday tech chat

PR #33, opened 2022-01-11
vb-438 log4j CVE-2021-45105

philip-milne:
Upgrade log4j version to 2.17.1 for gradle build


rj-adams (approved):
Looks fine - I'm assuming this is a problem for many projects - not just this one

philip-milne:
Thats right, it also looks to be in many cases a false negative, pushing the version up will hopefully keep the build happy while the underlying issue is resolved.

PR #34, opened 2022-01-12
DPS Gradle upgrade to 4.0.1

philip-milne:
DPS gradle plugin upgrade for combined fix:

- CVE-2021-44832
- CVE-2021-45105
- CVE-2021-42550


rj-adams (approved):
I don't think Lucy can do this now as her contract is with Syscon - an external supplier now supplying specific work items

PR #41, opened 2022-02-02
vb-450 expired reserved visit scheduled task

philip-milne:
VB-450 Reserve Visit Slot - Expired edge case.

Scheduled Task created to delete expired reservations if not modified within validity period.
Configurable cron and validity period added to application properties.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/task/VisitTask.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/task/VisitTask.kt
@@ -1,0 +1,44 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.task
+
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.visitscheduler.data.filter.VisitFilter
+import uk.gov.justice.digital.hmpps.visitscheduler.jpa.VisitStatus
+import uk.gov.justice.digital.hmpps.visitscheduler.service.VisitSchedulerService
+import java.time.LocalDateTime
+
+@Component
+class VisitTask(
+  private val visitSchedulerService: VisitSchedulerService,
+  @Value("\${task.expired-visit.enabled:false}") private val enabled: Boolean,
+  @Value("\${task.expired-visit.validity-minutes:20}") private val expiredPeriod: Long
+) {
+
+  @Scheduled(cron = "\${task.expired-visit.cron:0 0/15 * * * ?}")
+  fun deleteExpiredReservations() {
+    if (!enabled) {
+      return
+    }
+
+    val expired = visitSchedulerService.findVisitsByFilter(
+      VisitFilter(
+        status = VisitStatus.RESERVED,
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/task/VisitTask.kt
```diff
@@ -0,0 +1,44 @@
...
+import java.time.LocalDateTime
+
+@Component
+class VisitTask(
+  private val visitSchedulerService: VisitSchedulerService,
+  @Value("\${task.expired-visit.enabled:false}") private val enabled: Boolean,
+  @Value("\${task.expired-visit.validity-minutes:20}") private val expiredPeriod: Long
+) {
+
+  @Scheduled(cron = "\${task.expired-visit.cron:0 0/15 * * * ?}")
```
andrewrlee, line 19: so some slots might have a live for up to 30 minutes? is that right?

Clean up at 12:00
Booking is made 12:00:01
Clean up at 12:15  (not out of date)
Clean up at 12:30 (visit expired)

andrewrlee: oh, it's overiden per env

src/main/resources/application.yml
```diff
@@ -83,3 +83,9 @@ oauth:
 policy:
   minimum-booking-notice-period-days: 2
   maximum-booking-notice-period-days: 28
+
+task:
+  expired-visit:
+    enabled: true
+    cron: 0 0 * * * *
```
andrewrlee, line 90: runs on the hour, so booking could be up to 1:19:59 old

src/main/resources/db/migration/V1_0__create_tables.sql
```diff
@@ -10,7 +10,9 @@ CREATE TABLE visit
     visit_start     timestamp with time zone NOT NULL,
     visit_end       timestamp with time zone NOT NULL,
     status          VARCHAR(80)   NOT NULL,
-    reasonable_adjustments    text
+    reasonable_adjustments    text,
+    create_timestamp timestamp default current_timestamp,
+    modify_timestamp timestamp default current_timestamp
```
andrewrlee, line 15: worth adding an index?

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/task/VisitTask.kt
philip-milne: For now yes.  The actual schedule has not been defined and is configurable, It might be the case that we only want to clean up over night.  You are correct, to ensure expired reservations are cleaned out immediately or as close as possible you would need to trigger the event every minute - this may or may not have an impact on the DB performance depending on the number of available connections and number of records.   Once we have realistic metrics we can set the defaults to the appropriate levels which will not impact performance.

src/main/resources/application.yml
philip-milne: Running on the hour would be a good starting point to see how often this edge case occurs, this value will be refined from alpha.

src/main/resources/db/migration/V1_0__create_tables.sql
philip-milne: Good point.  I will create a backlog item to review the existing schema.

PR #50, opened 2022-02-28
VB-521 Visit Visitor Concerns text

philip-milne:
Adds the visitor concerns column to the visit table

Provide a field to store visitor concerns when migrating visit data.


andrewrlee (approved):
LGTM

PR #59, opened 2022-03-21
re-implement equals hash toString

philip-milne:
Prevents stack overflow in equals, hash & toString

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitContact.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitContact.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.visitscheduler.jpa
 
+import org.hibernate.Hibernate
 import javax.persistence.Column
 import javax.persistence.Entity
 import javax.persistence.Id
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitContact.kt
```diff
@@ -25,4 +26,27 @@ data class VisitContact(
...
+) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
+    other as VisitContact
+
+    if (id != other.id) return false
+    if (contactName != other.contactName) return false
+    if (contactPhone != other.contactPhone) return false
+
```
AledWEvans, line 38: Could be :
```
override fun equals(other: Any?): Boolean {
    if (this === other) return true
    if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
    
    other as VisitContact

    return (id == other.id) && (contactName == other.contactName) && (contactPhone == other.contactPhone)
}
```
or 

```
   if (id != other.id  || 
       contactName != other.contactName || 
       contactPhone != other.contactPhone) return false;
```

you call really!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitContact.kt
```diff
@@ -25,4 +26,27 @@ data class VisitContact(
...
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = id.hashCode()
+    result = 31 * result + contactName.hashCode()
```
AledWEvans, line 44: 31 could be constant

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitContact.kt
AledWEvans: I think ID compare is enough as it's a Primary key.

AledWEvans (approved):
looks much better

PR #68, opened 2022-03-31
VB-554 Comment Text attribute for a visit entry

AledWEvans:

Allows visit comment to be added to via visit Notes 

To bring in visit notes from the legacy system


philip-milne (approved):
LGTM

PR #70, opened 2022-04-01
VB-612 Refine Visit Reason status for legacy

AledWEvans:

Allows legacy status change reasons to be stored in a generic way in new system

To store legacy data in the new system without impacting the new system with legacy structures


philip-milne (approved):
LGTM

PR #74, opened 2022-04-04
VB-635 Remove sub status types from VisitStatus

AledWEvans:

remove unwanted status types

to remove unwanted code

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitStatus.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitStatus.kt
@@ -6,8 +6,6 @@ enum class VisitStatus(
 ) {
   RESERVED("Reserved"),
   BOOKED("Booked"),
-  CANCELLED_BY_PRISONER("Cancelled by Prisoner"),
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/jpa/VisitStatus.kt
```diff
@@ -6,8 +6,6 @@ enum class VisitStatus(
 ) {
   RESERVED("Reserved"),
   BOOKED("Booked"),
-  CANCELLED_BY_PRISONER("Cancelled by Prisoner"),
-  CANCELLED_BY_VISITOR("Cancelled by Visitor"),
-  CANCELLED_BY_PRISON("Cancelled by Prison"),
+  CANCELLED("Cancelled"),
   ATTENDED("Attended")
```
philip-milne, line 10: Do you think we need the status 'Attended' and if so should we also have 'Non-Attended'?  Do you think this would be better as an outcome as per the migrated data.

AledWEvans: Makes sense I suppose, did not really think of that!

Will add :

- Attended
- Non-Attended

philip-milne (approved):
LGTM

PR #77, opened 2022-04-06
VB-640 Refactor of package names

AledWEvans:

It reorganises the packages into standard conventions

to increase readability and knowledge


philip-milne (changes requested):
Much nicer.
File names need to match the class names.

PR #80, opened 2022-04-07
Vb 648 change booking reference to lower case

AledWEvans:

Changes the reference to lower case

improve memorability of reference for customers


philip-milne (approved):
LGTM

PR #82, opened 2022-04-08
VB-652 Rollback while circleci is not responding

philip-milne:
Rolls back the VB-652

CircleCI is not responding - so we are unable to test the deployment


rj-adams (approved):
never mind - worth a try

PR #85, opened 2022-04-19
VB-648 Defect in padding

philip-milne:
Corrects a defect in the padding causing repeated chars in the reference.

The reference needs to be random and must contain only one of the separator chars.  The refactor changed the use of the separator char pool to also be used to generate the random padding which resulted in repeated char sequences.


tpmcgowan (approved):
👍

PR #89, opened 2022-04-22
VB-661 Status query param added to get visits endpoint

philip-milne:
VisitController updated to include a status parameter on get visits endpoint

eg:
{{hostVisit}}/visits?visitStatus=BOOKED

Allows the client to filter visits by status


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
```diff
@@ -129,15 +130,21 @@ class VisitController(
...
   ): List<VisitDto> =
     visitService.findVisitsByFilter(
       VisitFilter(
         prisonerId = prisonerId?.trim(),
         prisonId = prisonId?.trim(),
         startDateTime = startTimestamp,
         endDateTime = endTimestamp,
-        nomisPersonId = nomisPersonId
+        nomisPersonId = nomisPersonId,
+        visitStatus = visitStatus
```
AledWEvans, line 147: arr so filter was already in

AledWEvans (approved):

PR #93, opened 2022-04-25
VB-667 Store the reason

AledWEvans:

The user can now store a reason and bespoke text against the canceled visit

To allow MOJ to monitor why visits are canceled

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -155,12 +155,12 @@ class VisitService(
 
     visitEntity.visitStatus = VisitStatus.CANCELLED
 
-    val statusChangeReason = createVisitNote(visitEntity, VISIT_OUTCOMES, cancelOutcome.outcome.name)
-    visitEntity.visitNotes.add(statusChangeReason)
+    val visitOutCome = createVisitNote(visitEntity, VISIT_OUTCOMES, cancelOutcome.outcome.name)
+    visitEntity.visitNotes.add(visitOutCome)
 
     cancelOutcome.text?.let {
-      val visitOutCome = createVisitNote(visitEntity, STATUS_CHANGED_REASON, cancelOutcome.text)
-      visitEntity.visitNotes.add(visitOutCome)
+      val statusChangeReason = createVisitNote(visitEntity, STATUS_CHANGED_REASON, cancelOutcome.text)
+      visitEntity.visitNotes.add(statusChangeReason)
     }
 
     visitRepository.saveAndFlush(visitEntity)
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -144,16 +147,23 @@ class VisitService(
...
+  fun cancelVisit(reference: String, cancelOutcome: OutcomeDto): VisitDto {
+    log.info("Updating visit for $reference with $cancelOutcome")
 
     val visitEntity = visitRepository.findByReference(reference)
     visitEntity ?: throw VisitNotFoundException("Visit reference $reference not found")
 
     visitEntity.visitStatus = VisitStatus.CANCELLED
 
-    // TODO: VB-667 Set Reason [VISIT_OUTCOMES, STATUS_CHANGED_REASON]
+    val visitOutCome = createVisitNote(visitEntity, VISIT_OUTCOMES, cancelOutcome.outcome.name)
```
philip-milne, line 158: Do we need something like this before this line to prevent 500 exceptions?

    if(visitEntity.visitNotes.any{it.type == VISIT_OUTCOMES})
      throw ValidationException("Visit ${visitEntity.reference} already contains OUTCOME")

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/CancelVisitTest.kt
```diff
@@ -0,0 +1,175 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.integration
```
philip-milne, line 1: So much better

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
AledWEvans: interesting point! have been wondering about this! what was the constraint,... If anything we should just remove an existing one,... theres only one out come right... it's been canceled!

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/CancelVisitTest.kt
AledWEvans: Yep going to try and slowly break up that big integration test class

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
AledWEvans: `    val visitOutCome = createVisitNote(visitEntity, VISIT_OUTCOMES, cancelOutcome.outcome.name)
    if (visitEntity.visitNotes.any { it.type == VISIT_OUTCOMES }) {
      log.warn("Removed old outcomes, as there is only one outcome when you cancel a visit!")
      visitEntity.visitNotes.removeAll { it.type == VISIT_OUTCOMES }
    }`

AledWEvans: Decided to let the normal constraint in the DB layer handle this,.. Otherwise we are adding layers!

PR #104, opened 2022-05-04
vb-638 domainevents helm config

philip-milne:
Adds namespace secrets to helm config for 'domainevents' topic


rj-adams (approved):
lets give it a try

PR #108, opened 2022-05-09
Vb 749 refactor current create endpoint

AledWEvans:

removed optional from val

makes it consistent with json anotations


philip-milne (approved):
LGTM

PR #109, opened 2022-05-09
VB-761 Remove DTO that is returned after creation/update of visit, re…

AledWEvans:
…turn reference instead

Does not return visit DTO on create and Update visit returns reference

We should only use the reference that is returned this code was originally written for testing purposes... but testing should never change code design.


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/CreateVisitTest.kt
```diff
@@ -176,7 +197,7 @@ class CreateVisitTest : IntegrationTestBase() {
     val jsonBody = BodyInserters.fromValue(migratedVisitRequestDto)
 
     // When
-    val responseSpec = callMigrateVisit(roleVisitSchedulerHttpHeaders, jsonBody)
+    val responseSpec = callCreateVisit(roleVisitSchedulerHttpHeaders, jsonBody)
```
AledWEvans, line 200: sorry this was a copy and past error from other change corrected them here!

PR #111, opened 2022-05-10
VB-784 Swagger UI documents a string is stated as required instead of…

AledWEvans:
… request json objects

Correct a bug with the swagger documentation on the api

to ensure the correct documentation is published to consumers of the api


tpmcgowan (approved):
👍

PR #116, opened 2022-05-13
Vb 691 fully support known mappings data

AledWEvans:

update to know mapping, legacy data is optional with lead visitor

make sure migration maps correctly


philip-milne (approved):
LGTM

PR #137, opened 2022-06-21
VB-830

AledWEvans:
Facility to enable/disable events being posted

toggel events that are sent Via SNS

allows us to test in live

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -12,6 +12,7 @@ generic-service:
     HMPPS_AUTH_URL: https://sign-in.hmpps.service.justice.gov.uk/auth
     PRISON_API_URL: https://api.prison.service.justice.gov.uk
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    FEATURE_EVENTS_SNS_ENABLED: false
 
 # CloudPlatform AlertManager receiver to route promethues alerts to slack
 # See https://user-guide.cloud-platform.service.justice.gov.uk/documentation/monitoring-an-app/how-to-create-alarms.html#creating-your-own-custom-alerts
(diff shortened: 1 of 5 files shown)


helm_deploy/values-prod.yaml
```diff
@@ -12,6 +12,7 @@ generic-service:
     HMPPS_AUTH_URL: https://sign-in.hmpps.service.justice.gov.uk/auth
     PRISON_API_URL: https://api.prison.service.justice.gov.uk
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    FEATURE_EVENTS_SNS_ENABLED: false
```
AledWEvans, line 15: Not sure what is required here @philip-milne and @rj-adams  what is the missing step here, surly I have to link this up in the cloud some how?

philip-milne (approved):
LGTM

helm_deploy/values-prod.yaml
rj-adams: If the values-prod.yaml defines the default value for the SNS switch in the prod environment, why is there not a similar line in the values.yaml which defines the default value for the SNS switch for other environments?

Or is there another mechanism which does this? and if so, why do we use different mechanisms for different environments?

AledWEvans: it defaults to true...

rj-adams: how does it default to true?

AledWEvans: @philip-milne informed me of the convention as follows :

feature:
  events:
    sns:
      enabled: true

is converted into the Eq of the below env variable

    FEATURE_EVENTS_SNS_ENABLED

which is then picked up by 

  @Value("\${feature.events.sns.enabled::true}")
  private val snsEventsEnabled: Boolean

rj-adams (dismissed):
LGTM

philip-milne (approved):
LGTM

PR #139, opened 2022-07-07
VB983/984 Enable events being posted pre prod/prod

AledWEvans:

Enable events by default being posted pre-prod and prod

We have tested the biss process flow to from end to end with out kicking off events, now we need to enable events after the test to allow the system to operate as exspected


rj-adams (approved):
LGTM

PR #140, opened 2022-07-15
Duplicates removed

philip-milne:
Removed duplicate dtos from the code base

DRY


AledWEvans (approved):
You could argue that these are different things, one is a create request and the other represents the thing you have created... but for now there is no difference and you could argue that you only separate them into two different things when you need to, therefore I approve..

PR #142, opened 2022-07-15
VB-1007 DayOfWeek-change-to-Session-Template-system

AledWEvans:

added days of week to session system

Improves scheduling of slots for VSiP


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/VisitSessionsControllerTest.kt
```diff
@@ -125,9 +78,10 @@ class VisitSessionsControllerTest : IntegrationTestBase() {
       .exchange()
       .expectStatus().isOk
       .expectBody()
-      .jsonPath("$.length()").isEqualTo(1)
-      .jsonPath("$[0].startTimestamp").isEqualTo("2021-01-08T09:00:00")
-      .jsonPath("$[0].endTimestamp").isEqualTo("2021-01-08T10:00:00")
+      .jsonPath("$.length()").isEqualTo(3)
+      .jsonPath("$[0].endTimestamp").isEqualTo("2021-01-10T10:00:00")
```
AledWEvans, line 82: need to sort this out

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/SessionTemplate.kt
```diff
@@ -52,4 +53,7 @@ data class SessionTemplate(
   @Column(nullable = false)
   val openCapacity: Int,
 
+  @Column(nullable = true)
```
philip-milne, line 56: Is nullable correct?

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/VisitSessionsControllerTest.kt
philip-milne, line 82: should this be 'startTimestamp'

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionServiceTest.kt
```diff
@@ -225,67 +165,6 @@ class SessionServiceTest {
       assertDate(sessions[0].startTimestamp, "2021-02-01T11:30:00", MONDAY)
     }
 
-    @Test
-    fun `all sessions are on past dates, no sessions are returned`() {
```
philip-milne, line 229: Is this no longer a valid test

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/SessionTemplate.kt
AledWEvans: depends, at this stage yes

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/VisitSessionsControllerTest.kt
AledWEvans: some of the tests need work,... I think the original was testing both start and end

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionServiceTest.kt
AledWEvans: not sure, I think I might have to still test for this!

philip-milne (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/SessionFrequency.kt
```diff
@@ -5,8 +5,6 @@ import java.time.Period
 enum class SessionFrequency(
```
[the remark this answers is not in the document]
rj-adams: I think frequency is now redundant in the database.

For a one-off session, the user can set the start date and expiry date to be close enough together to only enable one session - or the scheduler can do this behind the scenes if a new API endpoint is built to accept a session_template POST with type "one-off" and a single date parameter - or similar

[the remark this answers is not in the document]
AledWEvans: Was not sure about this, but will take it out as we have consensus on it I think!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/CreateSessionTemplateRequestDto.kt
```diff
@@ -38,4 +39,6 @@ data class CreateSessionTemplateRequestDto(
   @Schema(description = "open capacity", example = "50", required = true)
   @field:NotNull
   val openCapacity: Int,
+  @Schema(description = "day of week fpr visit", example = "MONDAY", required = false)
```
tpmcgowan, line 42: Typo here and in `SessionTemplateDto.kt` (`'fpr'`)

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/VisitSessionDto.kt
```diff
@@ -37,4 +58,7 @@ data class VisitSessionDto(
   @Schema(description = "The end timestamp for this visit session", example = "1", required = true)
   val endTimestamp: LocalDateTime,
 
+  @Schema(description = "The day of week for the visit session", example = "MONDAY", required = true)
```
tpmcgowan, line 61: Is `dayOfWeek` actually needed on this as for an individual visit session it should be what's in `startTimestamp`? Thinking that in the front-end when we retrieve the list of visit sessions to display we show day and date for each so would be using the timestamp rather than this.

AledWEvans: It is true, you don't need it, does add anything readability wise/ or for logging? your call?

philip-milne: Are individual visit sessions MVP?

rj-adams: I think the example value of 1 is misleading here - 1 converted to a timestamp is a few thousand years BC. It would be better to use a more realistic value - as a timestamp e.g. "23-May-2023 11:00:00" - in correct format of course.

AledWEvans: It is totally misleading.. for a start its not a timestamp... doc is wrong also,.. not really part of review but will change doc, but not the val name... that will have to wait

tpmcgowan: > It is true, you don't need it, does add anything readability wise/ or for logging? your call?

I'd say probably not worth it at the moment as - at least from front-end app point of view.

AledWEvans: Cool, I will take it out then!

tpmcgowan (approved):
LGTM!

philip-milne (approved):
LGTM

tpmcgowan (approved):
LGTM

rj-adams (approved):
LGTM

PR #143, opened 2022-07-21
VB-1007 DayOfWeek-change-to-Session-Template-system fix bug when depl…

AledWEvans:
deploying to environments fails because existing data does not conform to null value constraint 

deletes session data

prevent bug


philip-milne (approved):
LGTM

PR #144, opened 2022-07-21
VB-932 Capitalise legacy contact name

philip-milne:
Capitalise legacy contact data

Sanitises migrated data via API for defect VB-932

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/CreateLegacyContactOnVisitRequestDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/CreateLegacyContactOnVisitRequestDto.kt
@@ -6,18 +6,18 @@ import javax.validation.constraints.NotBlank
 
 data class CreateLegacyContactOnVisitRequestDto(@field:NotBlank val name: String, @field:NotBlank val telephone: String) {
   companion object {
-    const val UNKNOWN = "Unknown"
+    const val UNKNOWN_TOKEN = "UNKNOWN"
 
     // Deserialization kotlin data class issue when name and/or telephone  = json type of null defaults do not get set hence below
     // JsonCreator and JvmStatic code
     @JsonCreator
     @JvmStatic
     fun create(
-      @Schema(description = "Contact Name", example = "John Smith", defaultValue = UNKNOWN, required = false)
+      @Schema(description = "Contact Name", example = "John Smith", defaultValue = UNKNOWN_TOKEN, required = false)
       name: String? = null,
-      @Schema(description = "Contact Phone Number", example = "01234 567890", defaultValue = UNKNOWN, required = false)
+      @Schema(description = "Contact Phone Number", example = "01234 567890", defaultValue = UNKNOWN_TOKEN, required = false)
       telephone: String? = null
     ) =
-      CreateLegacyContactOnVisitRequestDto(name ?: UNKNOWN, telephone ?: UNKNOWN)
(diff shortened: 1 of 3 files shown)


psoleckimoj (dismissed):
LGTM!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/CreateLegacyContactOnVisitRequestDto.kt
```diff
@@ -6,7 +6,7 @@ import javax.validation.constraints.NotBlank
 
 data class CreateLegacyContactOnVisitRequestDto(@field:NotBlank val name: String, @field:NotBlank val telephone: String) {
   companion object {
-    private const val UNKNOWN = "UNKNOWN"
+    const val UNKNOWN = "Unknown"
```
AledWEvans, line 9: I'm not sure if this should be done, on server especially the entity because we would have migrated the data already! you could certainly change it in the DTO being returned. But I'm not rally sure if this is a concern of the entity or even the server side, it is a display concern.

AledWEvans (changes requested):
consider not doing this on the server side!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/CreateLegacyContactOnVisitRequestDto.kt
AledWEvans: If you do this, you would have to update the entries in the DB!

philip-milne: There is defiantly an inconstancy in how the UI is representing this data - im not sure the UI should be manipulating this text.  The api should be omitting unknown fields from the response, but these fields are NOT NULL & NOT BLANK for VSiP so we decided to populate migrated records with an unknown text value.  I believe the UI should display the text as it receives it in this case, otherwise were have code in the UI to handle a temporary fix for migrated records.

With the existing migrated data, we only have dev and test migration data at present which we can drop and re-migrate or leave and migrate some new records for testing this change.  As far as im aware nothing has been migrated yet in the PROD environment.

philip-milne: Ignoring the suggested fix on the defect.  Another solution to this issue is that the UI simply displays the text as is which would be all uppercase for unknown contact data.  The assumption in the defect from planning is that the capitalisation of the contact name is correct and the phone number is wrong but we can easily say the reverse which means leaving the API as is and updating the UI to remove the manipulation of the contact name when the name is unknown - this would still need a change in the UI anyway.

AledWEvans: I think the last suggestion is correct.. it's a UI change as it is a UI concern as far as the server side is concerned it's fine... we should not have inconsistency in the DB

philip-milne:
Updated to capitalise migrated contact names when provided in uppercase.

philip-milne:
Requires dropping or ignoring existing migrated data for testing in DEV & PRE-PROD

psoleckimoj (approved):
LGTM!

AledWEvans (approved):
Nice work, much better solultion

PR #145, opened 2022-07-22
VB-653

AledWEvans:
VB-653 The scheduler should only accept visits of type "Social"

removes visit types that are not currently support by the system.

To make sure code follows the requirements


philip-milne (approved):
LGTM

PR #155, opened 2022-09-02
VB-1140 Visit scheduler timing out

AledWEvans:

attempts to fix timeout by reducing calls to DB with checking non-association with visits

to fix the bug

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
@@ -1,12 +1,13 @@
 package uk.gov.justice.digital.hmpps.visitscheduler.service
 
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.http.HttpStatus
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -44,6 +45,10 @@ class SessionService(
   private val policyNonAssociationWholeDay: Boolean,
 ) {
 
+  companion object {
+    val LOG: Logger = LoggerFactory.getLogger(this::class.java)
```
dhirajshettymoj, line 49: Minor observation, we could rename this val to log to keep it consistent with other instances.

AledWEvans: Not sure as it is a constant and the convention with constants is uppercase right?

dhirajshettymoj: Ok, yes as mentioned a very minor observation.

dhirajshettymoj:
LGTM!!

dhirajshettymoj (approved):
LGTM!!

PR #156, opened 2022-09-07
VB-1157-Defect---correct-R__VB-819_add_hewell_code_HEI_prison_data.sql

AledWEvans:

corrects a case issue with the main room name from **visits main room** to **Visits Main Room**

to match VSiP with other services


psoleckimoj:
@AledWEvans I wonder if this is correct now since there is no main visits room, we just have Open, Closed and Unknown on the visits page now.

AledWEvans:
It's a funny one really! is still used in migration and we were getting different cases, so all this PR does is make the case consistent, see https://dsdmoj.atlassian.net/browse/SDI-224. We should perhaps remove it later...

psoleckimoj (approved):
LGTM!

PR #158, opened 2022-09-08
VB-1168 Add new get end point /visits/booked/{reference}

AledWEvans:

fixes bug desribed in VB-1166

To fix the bug

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
@@ -18,6 +18,12 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
   fun findByReference(reference: String): Visit?
 
   @Query(
+    "SELECT v FROM Visit v " +
+      "WHERE v.reference = :reference and v.visitStatus = 'BOOKED'"
+  )
+  fun findBookedVisitByReference(reference: String): Visit?
+
+  @Query(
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -17,6 +17,12 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
 
   fun findByReference(reference: String): Visit?
 
+  @Query(
+    "SELECT v FROM Visit v " +
+      "WHERE v.reference = :reference and v.visitStatus = 'BOOKED'"
+  )
+  fun findBookedVisitByReference(reference: String): Visit?
```
dhirajshettymoj, line 24: Would there be value in passing the status as a parameter or as a List of VisitStatuses and doing an IN in the query?

AledWEvans: I tend to only put features or logic in when I need it rather than if I need it!

dhirajshettymoj (approved):
Looks good.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
rj-adams: I'm not convinced the URL is constructed correctly for this

I would expect it to be either:

/visits/{visit reference}/booked
or
/visits/{visit reference}?status=booked

PR #159, opened 2022-09-09
Vb 1166 the UI allows me to cancel twice and when i do i get a http 500 error

AledWEvans:

Removal of end point code for booked visits as another solution has been found

To tidy up code and prevent unused code being available


dhirajshettymoj (approved):
Approved.

PR #160, opened 2022-09-09
VB-1168 revert of code as get booked visits no longer needed

AledWEvans:

reverting code that is no longer needed

to keep code base clean


dhirajshettymoj (approved):
Compared against #158 and can see the changes have been regressed.

PR #161, opened 2022-09-09
WIP - VB-1171 - performance improvement to populateBookedCount method.

dhirajshettymoj:
Performance improvements around SQL calls in populateBookedCount. Existing logic to do a VisitRepository.findAll(VisitSpecification) has been replaced with select count(*) and grouping by VisitRestrictions to reduce number of calls made to DB.

Reduce the number of calls to DB as part of /visit-sessions to help improve performance (response times) similar to. changes done for VB-1140.


dhirajshettymoj:
To summarise the changes  - 
 The existing call to do a findAll(...) by visitspecification to get all visits and then do count by STATUS has been removed as it unnecessarily gets Visit and VisitContact data while what we need is a count.
 There will be a single SQL call for both OPEN and CLOSED restrictions - per session. the call is a count SQL with  a group by VisitRestriction. There is further logic in Service layer to get the counts by restriction (OPEN or CLOSED).

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -31,4 +33,22 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
...
   ): Boolean
+
+  @Query(
+    "SELECT v.visitRestriction as visitRestriction, count(v) as count  FROM Visit v " +
+      "WHERE v.prisonId = :prisonId and " +
+      "v.visitStart >= :startDateTime and " +
+      "v.visitStart < :endDateTime and " +
+      "v.visitRoom = :visitRoom and " +
+      "(v.visitRestriction = 'OPEN' or v.visitRestriction = 'CLOSED') and " +
+      " v.visitStatus in (:visitStatuses) " +
```
AledWEvans, line 44: I see why you have done this, if you have change code to use a visitStatuses, I would refactor my other code (line 28) above to use the same approach to be constant other wise we have two ways of doing the same thing. Consistency is very important, might not be a problem with a one off but when you get a 100 one offs over time it becomes a problem

dhirajshettymoj: Fair point. I will update my code. I think having the statuses in the SQL makes it more readable and its only 2 statuses so IN does not make a lot of sense.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -47,6 +48,7 @@ class SessionService(
 
   companion object {
     val LOG: Logger = LoggerFactory.getLogger(this::class.java)
+    val ACTIVE_VISIT_STATUSES: List<VisitStatus> = listOf(VisitStatus.BOOKED, VisitStatus.RESERVED)
```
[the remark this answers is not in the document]
dhirajshettymoj: This list will no longer be needed after the below change. So I will revert it back.

AledWEvans (approved):
Nice work

AledWEvans (changes requested):
tested failed

AledWEvans (dismissed):
I re-ran tests and it worked this time,... must be a server issue

AledWEvans (approved):

AledWEvans (approved):

PR #165, opened 2022-09-14
Vb 1197 WIP - reduce number of calls to prison api in get visit sessions

dhirajshettymoj:
Make only 1 call to prison API to get the offender's list for a prisonerId. Currently the calls are being multiple times i.e. for each session.

Performance improvements and reduced downstream calls.

WIP change only, not to be merged now.


AledWEvans (approved):
LGTM

PR #167, opened 2022-09-22
VB-1206 amend endpoints to new structure, VB-1209 Keep old endpoints

AledWEvans:

This pull request allows for multiple visits to share the same reference, it also add new end point to allow for this to be controlled correctly and keeps the old ones.

to allow the customer to keep there reference across phone calls for an intended visit


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -49,4 +51,12 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
...
     endDateTime: LocalDateTime
   ): List<VisitRestrictionStats>
+
+  @Query(
+    "SELECT v FROM Visit v WHERE v.reference = :reference and v.visitStatus = 'BOOKED' "
+  )
+  fun findBookedVisit(reference: String): Visit?
+
+  @Query("SELECT * FROM visit WHERE reference = :reference AND visit_status = 'RESERVED' ORDER BY modify_timestamp LIMIT 1", nativeQuery = true)
+  fun findReservedVisit(reference: String): Visit?
```
AledWEvans, line 61: Need to add a test for this! The order side of things

AledWEvans:
After discussion with @dhirajshettymoj and @tpmcgowan we will introduce a application reference.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
```diff
@@ -22,35 +21,131 @@ import org.springframework.web.bind.annotation.RequestParam
...
+      ),
+      ApiResponse(
+        responseCode = "404",
+        description = "Visit slot not found",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))]
+      ),
+    ]
+  )
+  fun changeReservedVisitSlot(
+    @Schema(description = "reference", example = "v9-d7-ed-7u", required = false)
```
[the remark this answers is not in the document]
AledWEvans: Good spot

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/ChangeReservedVisitSlotRequestDto.kt
```diff
@@ -0,0 +1,26 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.visitscheduler.controller.validators.VisitorContactValidation
+import uk.gov.justice.digital.hmpps.visitscheduler.model.VisitRestriction
+import uk.gov.justice.digital.hmpps.visitscheduler.validation.NullableNotEmpty
+import java.time.LocalDateTime
+import javax.validation.Valid
+
+data class ChangeReservedVisitSlotRequestDto(
```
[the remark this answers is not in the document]
AledWEvans: They have common data elements but not common in type e.g. we are talking here about apples and oranges not apples and fruit..

rj-adams:
calling it an "application" reference feels like it might be confusing - a "visitRequest" reference maybe?

AledWEvans:
I like applicationReference actually, say what it is, also added a slightly different format from booking reference so we can spot the difference

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -49,4 +53,12 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
     startDateTime: LocalDateTime,
     endDateTime: LocalDateTime
   ): List<VisitRestrictionStats>
+
+  @Query(
+    "SELECT v FROM Visit v WHERE v.reference = :reference and v.visitStatus = 'BOOKED' "
+  )
+  fun findBookedVisit(reference: String): Visit?
+
+  @Query("SELECT * FROM visit WHERE reference = :reference AND visit_status = 'RESERVED' ORDER BY modify_timestamp DESC LIMIT 1 ", nativeQuery = true)
```
dhirajshettymoj, line 62: This method might no longer be needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -158,21 +253,6 @@ class VisitService(
...
-        "visit-scheduler-prison-visit-deleted",
-        mapOf(
-          "reference" to reference
-        ),
-        null
-      )
-    } ?: run { log.debug("Visit reference $reference not found") }
-  }
-
   fun deleteAllVisits(visits: List<VisitDto>) {
```
dhirajshettymoj, line 256: This method needs to take the application reference instead of reference when called from the delete task?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -33,8 +36,100 @@ class VisitService(
...
+        if (changeReservedVisitSlotRequestDto.startTimestamp != null) "visitStart" to changeReservedVisitSlotRequestDto.startTimestamp.format(DateTimeFormatter.ISO_DATE_TIME) else null,
+        if (visitEntity.visitStatus != null) "visitStatus" to visitEntity.visitStatus.name else null
+      ).toMap(),
+      null
+    )
+
+    return VisitDto(visitEntity)
+  }
+
+  @Suppress("KotlinDeprecation")
```
dhirajshettymoj, line 132: Should we add @Deprecated annotation here if being called only from LegacyController?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -18,6 +18,10 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
 
   fun findByReference(reference: String): Visit?
 
+  fun findByApplicationReference(applicationReference: String): Visit?
```
dhirajshettymoj, line 21: deleteAllByReferenceIn needs to take an application reference instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
AledWEvans: True, I think it's used in a test so will remove method and changed test

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
AledWEvans: There is another task for this not part of this review

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
AledWEvans: There is another task for this not part of this review

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
AledWEvans: @Suppress("KotlinDeprecation") is the same thing!

dhirajshettymoj (approved):
Looks good to me.

PR #168, opened 2022-09-23
VB-1225 `applicationReference` missing on existing bookings

AledWEvans:

fixes `applicationReference` missing on existing bookings

to fix a bug, other wise old visits will not be avalable

### DIFF
--- a/src/main/resources/db/migration/V2_05__fix_missing_application_reference_defect.sql
+++ b/src/main/resources/db/migration/V2_05__fix_missing_application_reference_defect.sql
@@ -1,0 +1,2 @@
+UPDATE visit SET application_reference = reference
+WHERE application_reference IS NULL OR  LENGTH(application_reference) = 0;


src/main/resources/db/migration/V2_05__fix_missing_application_reference_defect.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE visit SET application_reference = reference
```
tpmcgowan, line 1: Assuming it doesn't matter that these application references won't have the new format?

tpmcgowan (approved):
LGTM - with caveat that I made as a comment

src/main/resources/db/migration/V2_05__fix_missing_application_reference_defect.sql
AledWEvans: Well the application reference is the same as the reference as it's a one to one, but it does not really matter that they are in a different format the new format

PR #170, opened 2022-09-27
VB-1213  Amending to solve booking count issue

AledWEvans:

It reduces book count issues, when previously booked visit is being updated

to give the user more accurate information in regards to slots that have been booked

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -10,7 +10,7 @@ import org.springframework.data.domain.Pageable
 import org.springframework.data.domain.Sort
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
-import uk.gov.justice.digital.hmpps.visitscheduler.dto.ChangeReservedVisitSlotRequestDto
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.ChangeVisitSlotRequestDto
 import uk.gov.justice.digital.hmpps.visitscheduler.dto.CreateVisitRequestDto
 import uk.gov.justice.digital.hmpps.visitscheduler.dto.OutcomeDto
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -44,50 +44,79 @@ class VisitService(
...
+      visitorSupport = reserveVisitSlotDto.visitorSupport
+    )
+  }
 
-    return VisitDto(visitEntity)
+  private fun isVisitNew(bookedVisit: Visit?, reserveVisitSlotDto: ReserveVisitSlotDto): Boolean {
+
+    if (bookedVisit == null ||
+      bookedVisit.prisonId != reserveVisitSlotDto.prisonId ||
+      bookedVisit.prisonerId != reserveVisitSlotDto.prisonerId ||
```
dhirajshettymoj, line 103: Didn't realise == calls equals in Kotlin.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -44,50 +44,79 @@ class VisitService(
...
+          }
+          visitEntity.support.add(createVisitSupport(visitEntity, it.type, it.text))
         }
-        visitEntity.support.add(createVisitSupport(visitEntity, it.type, it.text))
       }
+
+      createVisitTrackEvent(visitEntity)
+      return VisitDto(visitEntity)
+    } else {
+      return changeVisitSlot(bookedVisit!!.applicationReference, mapToChangeVisitSlotRequestDto(reserveVisitSlotDto))
```
dhirajshettymoj, line 86: Presuming bookedVisit will be not null as we have done the check in if clause. Is that right?

AledWEvans: It's never null at this stage

dhirajshettymoj (approved):

PR #171, opened 2022-09-27
VB-1238 Fix dependancy issue changed : uk.gov.justice.hmpps.gradle-sp…

AledWEvans:
…ring-boot to 4.5.1

makes sure we are using the latest common dependancies

_Required._

security

_Required._


tpmcgowan (approved):
LGTM

PR #172, opened 2022-09-27
VB-1213 Amending to solve booking count issue

AledWEvans:
New approach reverted changes from this morning,
now uses status

it makes the booking count more accurate and does not mess up the confirm flow 

to help the use know how many slots are booked

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
@@ -175,7 +175,7 @@ class VisitController(
     @PathVariable reference: String,
     @RequestBody @Valid reserveVisitSlotDto: ReserveVisitSlotDto
   ): VisitDto {
-    return visitService.reserveVisitSlot(reference.trim(), reserveVisitSlotDto)
+    return visitService.changeBookedVisit(reference.trim(), reserveVisitSlotDto)
   }
 
   @PreAuthorize("hasRole('VISIT_SCHEDULER')")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
@@ -50,7 +50,7 @@ class VisitController(
   @PostMapping(VISIT_RESERVE_SLOT)
   @ResponseStatus(HttpStatus.CREATED)
   @Operation(
-    summary = "Reserve a slot (date/time slot) for a visit ",
+    summary = "Reserve a slot (date/time slot) for a visit (a starting point)",
     requestBody = io.swagger.v3.oas.annotations.parameters.RequestBody(
       content = [
         Content(
@@ -91,7 +91,7 @@ class VisitController(
   @PutMapping(VISIT_RESERVED_SLOT_CHANGE)
   @ResponseStatus(HttpStatus.OK)
   @Operation(
-    summary = "Change a reserved slot and associated details for a visit ",
+    summary = "Change a reserved slot and associated details for a visit (before booking)",
     requestBody = io.swagger.v3.oas.annotations.parameters.RequestBody(
       content = [
         Content(
@@ -139,7 +139,7 @@ class VisitController(
   @PutMapping(VISIT_CHANGE)
   @ResponseStatus(HttpStatus.CREATED)
   @Operation(
-    summary = "Change a booked visit",
+    summary = "Change a booked visit, (a starting point)",
     requestBody = io.swagger.v3.oas.annotations.parameters.RequestBody(
       content = [
         Content(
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
```diff
@@ -91,7 +91,7 @@ class VisitController(
   @PutMapping(VISIT_RESERVED_SLOT_CHANGE)
   @ResponseStatus(HttpStatus.OK)
   @Operation(
-    summary = "Change a reserved slot and associated details for a visit ",
+    summary = "Change a reserved slot and associated details for a visit (before booking)",
```
tpmcgowan, line 94: @AledWEvans as discussed just now, reminder that this needs to cope with visit statuses that are `RESERVED` or `CHANGING`. (Probably worth tweaking the summary text there too so that's clear in the documentation)

AledWEvans: Yes will make change asap to change status during the VISIT_RESERVED_SLOT_CHANGE call if need be

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -43,77 +44,71 @@ class VisitService(
...
   }
 
   fun changeVisitSlot(applicationReference: String, changeVisitSlotRequestDto: ChangeVisitSlotRequestDto): VisitDto {
     val visitEntity = visitRepository.findByApplicationReference(applicationReference) ?: throw VisitNotFoundException("Reserved visit reference $applicationReference not found")
 
     changeVisitSlotRequestDto.visitRestriction?.let { visitRestriction -> visitEntity.visitRestriction = visitRestriction }
-    changeVisitSlotRequestDto.startTimestamp?.let { visitStart -> visitEntity.visitStart = visitStart }
+    changeVisitSlotRequestDto.startTimestamp?.let {
+      visitEntity.visitStart = it
+      visitEntity.visitStatus = getStartingStatus(visitEntity.reference, startTimestamp = changeVisitSlotRequestDto.startTimestamp)
```
dhirajshettymoj, line 110: just a query, do we need this outside the loop?

AledWEvans: I did think about this, the only thing that can change status at this stage is the startTimestamp if it's null it has to change since the step so no status needs to be updates,.. if it has change the status might need to change.. I think we are good

dhirajshettymoj (dismissed):

tpmcgowan:
I've noticed that calling `PUT /visits/{reference}/change` for a visit reference that **doesn't** exist actually creates a reservation. Is this OK or should it only work with BOOKED visits? (Not causing a problem but perhaps unexpected behaviour 🤔)

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -43,77 +44,71 @@ class VisitService(
...
-  private fun isVisitNew(bookedVisit: Visit?, reserveVisitSlotDto: ReserveVisitSlotDto): Boolean {
+  private fun getStartingStatus(bookingReference: String, prisonId: String? = null, prisonerId: String? = null, startTimestamp: LocalDateTime): VisitStatus {
+
+    val bookedVisit = this.visitRepository.findBookedVisit(bookingReference)
 
     if (bookedVisit == null ||
-      bookedVisit.prisonId != reserveVisitSlotDto.prisonId ||
-      bookedVisit.prisonerId != reserveVisitSlotDto.prisonerId ||
-      bookedVisit.visitStart.compareTo(reserveVisitSlotDto.startTimestamp) != 0
+      (prisonId != null && bookedVisit.prisonId != prisonId) ||
```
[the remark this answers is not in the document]
tpmcgowan: Likewise, for `PUT /visits/{applicationReference}/slot/change` if you change the `startTimestamp` it'll go from `CHANGED` to `RESERVED` (as expected) but if you change `visitRestriction` it'll stay as `CHANGED` - as above I think this isn't right?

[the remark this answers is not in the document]
AledWEvans: Yes will look into that, it currently does not case about visitRestriction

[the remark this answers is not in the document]
AledWEvans: added Restrictions

dhirajshettymoj (approved):
Looks good (I presume this fixes the issue Tom mentioned)

tpmcgowan (approved):
LGTM - fixes the issues I raised with `visitRestriction` and unexpectedly creating new visits

PR #173, opened 2022-09-28
VB-1242 Using getBookedVisit on /visits/{reference}

AledWEvans:

Fixes a defect VB-1242 - Using getBookedVisit on /visits/{reference} should be allowing to get booked and canceled visits

to allow the cancellation sync code to get canceled Visits


andymarke (approved):
LGTM

PR #174, opened 2022-09-29
VB-1249 500 error on GET /visits/{reference}

AledWEvans:

Basically more than one visit was being return in a DB call that only expected one. now we return the latest BOOKED or CANCELED visit

to fix bug


tpmcgowan (approved):
Tested and works as expected 👍 (will leave @dhirajshettymoj to comment on the Kotlin!)

dhirajshettymoj (approved):
LGTM

PR #176, opened 2022-10-03
Feature/vb 1246 update visit scheduler events sent to application insights

dhirajshettymoj:

Changes to send updated events to application insights given the change in API calls.

Updated code to send new and updated custom events to application insights which will eventually be used to create reports.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/TelemetryVisitEvents.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/TelemetryVisitEvents.kt
@@ -1,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.service
+
+enum class TelemetryVisitEvents(val eventName: String) {
+  VISIT_SLOT_RESERVED_EVENT("visit-slot-reserved"),
+  VISIT_CHANGED_EVENT("visit-changed"),
+  VISIT_BOOKED_EVENT("visit-booked"),
+  VISIT_SLOT_CHANGED_EVENT("visit-slot-changed"),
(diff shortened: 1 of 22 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/TelemetryVisitEvents.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.service
```
AledWEvans, line 1: good call

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -351,14 +328,14 @@ class VisitService(
...
         "reference" to visitToBook.reference,
         "visitStatus" to visit.visitStatus.name,
-      ).toMap(),
-      null
+        "applicationReference" to visit.applicationReference,
+        "isUpdated" to changedVisit.toString()
+      ).toMap()
     )
 
     if (changedVisit) {
```
AledWEvans, line 341: Could possible have an event to say a booked visit has changed... but could add that later when we need it rather than if

AledWEvans (approved):
Nice work will make insights way more useful..

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
dhirajshettymoj: Yes that was the plan earlier. Changed it based on the Friday call. With this approach the event will closely match the API call.

PR #179, opened 2022-10-06
VB-1263 Add creation date and update date to migration api

AledWEvans:

allows nomis to send create and update date to be passed into the VSiP system 

Allows users of visip to know when the visit was created rather than the migrated time in VSIP


AledWEvans:
I have put this to WIP as we have not had a convo with SYSCON yet

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -100,4 +102,20 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
     "SELECT CASE WHEN (COUNT(v) = 1) THEN TRUE ELSE FALSE END  FROM Visit v WHERE v.reference = :bookingReference AND v.visitStatus = 'BOOKED' "
   )
   fun isValidBookingReference(bookingReference: String): Boolean
+
+  @Transactional
+  @Modifying
+  @Query(
+    "Update visit SET  modify_timestamp = :modifyTimestamp WHERE id=:id",
+    nativeQuery = true
```
dhirajshettymoj, line 110: We could change this from nativeQuery to a JPA query.

AledWEvans: native could be much faster!

dhirajshettymoj (approved):
Looks good to me.

PR #180, opened 2022-10-07
Feature/vb 1253 server side prevent cancellation or amendment after visit start time

dhirajshettymoj:
VB-1253 changes - do not allow a visit to be amended or cancelled if the visit has expired i.e. the visit start date is in the past. Checks added to the following API calls -

PUT /visits/{reference}/change
PUT /visits/{reference}/cancel
PATCH /visits/{reference}/cancel
PUT /visits/{applicationReference}/book (if updating an existing visit)

Fix for VB-1253, to avoid expired visits from being amended.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -33,6 +33,7 @@ import uk.gov.justice.digital.hmpps.visitscheduler.repository.VisitRepository
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
 import java.util.function.Supplier
+import javax.validation.ValidationException
 
 @Service
 @Transactional
@@ -51,10 +52,14 @@ class VisitService(
   }
 
   fun changeBookedVisit(bookingReference: String, reserveVisitSlotDto: ReserveVisitSlotDto): VisitDto {
+    val visit = visitRepository.findBookedVisit(bookingReference)
+      ?: throw VisitNotFoundException("Visit booking reference $bookingReference not found")
 
-    if (!visitRepository.isValidBookingReference(bookingReference)) {
-      throw VisitNotFoundException("Visit booking reference $bookingReference not found")
-    }
+    // check if the visit is in the past
+    if (isBookingStartDateInPast(visit)) throw ExpiredVisitAmendException(
+      "Visit with booking reference - $bookingReference is in the past, it cannot be changed",
+      ExpiredVisitAmendException("trying to change an expired visit")
+    )
 
     return reserveVisitSlot(bookingReference, reserveVisitSlotDto)
   }
@@ -229,7 +234,7 @@ class VisitService(
   @Transactional(readOnly = true)
   fun findExpiredApplicationReferences(): List<String> {
     val localDateTime = getReservedExpiredDateAndTime()
-    log.debug("Entered findExpiredApplicationReferences :" + localDateTime)
+    log.debug("Entered findExpiredApplicationReferences : $localDateTime")
     return visitRepository.findExpiredApplicationReferences(localDateTime)
   }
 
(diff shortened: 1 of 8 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/helper/EntityDataLoader.kt
```diff
@@ -29,7 +29,7 @@ class VisitEntityHelper(
     prisonerId: String = "FF0000AA",
     prisonId: String = "MDI",
     visitRoom: String = "A1",
-    visitStart: LocalDateTime = LocalDateTime.of(2021, 11, 1, 12, 30, 44),
+    visitStart: LocalDateTime = LocalDateTime.of((LocalDateTime.now().year + 1), 11, 1, 12, 30, 44),
```
AledWEvans, line 32: I think you could change this to LocalDateTime.now(),...

Realised the testing was on  telemetry assert, I think it's because it uses ISO_DATE_TIME so instead of truncating it I would advise the following or something similar :

assertThat(it["visitStart"]).isEqualTo(reservedVisit.visitStart.format(DateTimeFormatter.ISO_DATE_TIME))

What ever the case I think the above assert is better way of testing this code

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -350,6 +365,13 @@ class VisitService(
   fun cancelVisit(reference: String, cancelOutcome: OutcomeDto): VisitDto {
     val visitEntity = visitRepository.findBookedVisit(reference) ?: throw VisitNotFoundException("Visit $reference not found")
 
+    // check if the visit being cancelled is in the past
+    if (isBookingStartDateInPast(visitEntity))
+      throw ExpiredVisitAmendException(
+        "Visit with booking reference - ${visitEntity.reference} is in the past, it cannot be cancelled",
+        ExpiredVisitAmendException("trying to cancel an expired visit")
+      )
+
```
AledWEvans, line 374: I always put the {} with my if's your call but do like things to be consistant

if (isBookingStartDateInPast(visitEntity)) {
       throw ExpiredVisitAmendException(
        "Visit with booking reference - ${visitEntity.reference} is in the past, it cannot be cancelled",
        ExpiredVisitAmendException("trying to cancel an expired visit")
      )
}

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -456,3 +482,11 @@ class SupportNotFoundException(message: String? = null, cause: Throwable? = null
     return SupportNotFoundException(message, cause)
   }
 }
+class ExpiredVisitAmendException(message: String? = null, cause: Throwable? = null) :
```
AledWEvans, line 485: do we put our exceptions in side the class, what convention are we using on this project

AledWEvans (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -51,10 +52,14 @@ class VisitService(
   }
 
   fun changeBookedVisit(bookingReference: String, reserveVisitSlotDto: ReserveVisitSlotDto): VisitDto {
+    val visit = visitRepository.findBookedVisit(bookingReference)
```
AledWEvans, line 55: Could create function call validateStart time, and reuse for the 3 cases

AledWEvans (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/helper/EntityDataLoader.kt
dhirajshettymoj: Updated the asserts for visitStart sent to telemetry client as above.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
dhirajshettymoj: Agree, updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
dhirajshettymoj: So far all exceptions are being added this way (in VisitService and SessionTemplateService).

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
dhirajshettymoj: Created validateVisitStartDate and moved the validation code there.

AledWEvans (approved):
Nice work LGTM

PR #182, opened 2022-10-16
Start Wiremock on a random port to avoid conflicts

bsideup:

This PR changes the test setup to use a random port when WireMock is started. Otherwise, the tests won't run if the machine that runs them already has something running on port `8092`.

Your friend from the [Testcontainers](https://github.com/testcontainers/testcontainers-java) project was looking at how the library is used in the wild, tried running tests and discovered the static port usage :)


dhirajshettymoj:
Thanks @bsideup for having a look at the code and the code review.

PR #185, opened 2022-10-18
VB-1270 remove clock dependency injected in session service

dhirajshettymoj:

Remove the dependency of injecting a clock for the tests to work. Instead updated the tests to ensure it works against the current date.

Code improvements.


AledWEvans (approved):
LGTM

PR #186, opened 2022-10-19
VB-1329 Remove 'Create/Delete session template' via endpoint

AledWEvans:

Remove code that is not used and removes code without requirements

to tidy up code base, code should not written before it is needed


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionTemplateController.kt
```diff
@@ -118,33 +72,4 @@ class VisitSessionTemplateController(
     @Schema(description = "Template id", example = "45645", required = true)
     @PathVariable templateId: Long
   ): SessionTemplateDto = sessionTemplateService.getSessionTemplates(templateId)
-
```
AledWEvans, line 121: @tpmcgowan Can we remove the delete end point also?

dhirajshettymoj: @AledWEvans what if we do not remove the endpoint instead throw an Exception every time it's called as these endpoints will be needed in a while?

AledWEvans: Well thats kind of the hole point is to remove it,... it's code that is based on NO requirements, if we want to revert it at a later date we can go back in git history and restore it! The reason why I want it removed is it's muddies the water and can involve refactoring it, even when we don't use it. It's actually simple code also so would just take us no time at all to re-write it.

dhirajshettymoj (approved):
Looks good.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionTemplateController.kt
tpmcgowan: > @tpmcgowan Can we remove the delete end point also? could not spot it in the UI code!

Yeah, don't see why not. We don't do anything with session templates in the front-end; we only query the available slots that are derived from them.

We've used these endpoints occasionally to create/delete specific template scenarios for testing but we can do that directly in the DB locally, for example, when needed.

rj-adams:
I think we should create a new role - Visit-Scheduler-Admin / VSIP-Admin - or similar which would secure the endpoint / UI.

Some new tickets required for this but the business side of VSIP won't be concerned with this - it's down to the teckie team to decide what they want to do - maybe a discussion topic for this Friday...

AledWEvans:
@rj-adams 
I'm assuming we are going to have an epic for admin, I was listening to the convo in stand up and after party this am and thought,... we need to start thinking about admin and how this will work. Thats when we introduce these roles... when we know a lot more about it.

PR #187, opened 2022-10-20
VB-945 Add prisoner wing to session template

AledWEvans:

Add prisoner wing to session template

To allow configuration data of the prisoner wing to be configured


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/prison/PrisonWing.kt
```diff
@@ -0,0 +1,30 @@
...
+import javax.persistence.Entity
+import javax.persistence.GeneratedValue
+import javax.persistence.GenerationType
+import javax.persistence.Id
+import javax.persistence.ManyToMany
+import javax.persistence.Table
+
+@Entity
+@Table(name = "PRISON_WING")
+data class PrisonWing(
```
AledWEvans, line 14: may be this should be SessionWing or AllowedSessionWing

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/SessionTemplateDto.kt
```diff
@@ -34,6 +34,8 @@ data class SessionTemplateDto(
   val openCapacity: Int,
   @Schema(description = "day of week fpr visit", example = "MONDAY", required = false)
   val dayOfWeek: DayOfWeek?,
+  @Schema(description = "List of prison wings associated with the session template", required = false)
```
dhirajshettymoj, line 37: If there are none associated this session is available to the whole prison, right? Should we mention this on the comments?

rj-adams: Can I just confirm that there can be multiple wings which can be associated with the session template entry. So if Wings D & F could have visits at the same time, the PrisonWings value would hold "DF" or similar

AledWEvans: Yes that's correct @rj-adams

AledWEvans: @dhirajshettymoj Yes I think it should mention it I will add

dhirajshettymoj (approved):
Looks good.

PR #189, opened 2022-10-24
VB-1312 - Update cron for deleteTask job to run every 10 minutes.

dhirajshettymoj:

 Update to run job every 10 minutes instead of an hour to delete any expired visits more than 20 (set in validity-minutes) minutes old.

Ensure expired visits are deleted at the earliest so that the slot is freed.


AledWEvans (approved):
LGTM

PR #190, opened 2022-10-24
VB-1339 Fixup some of the get session tests

AledWEvans:

Fixed some get session tests, as they may have failed depending on when they are run

To improve testing reliability


dhirajshettymoj (approved):
Looks good to me.

PR #194, opened 2022-10-26
Vb 1339 fixup some of the get session tests

AledWEvans:

Tests were failing during certain times of week, also one other test was randomly failing (probably due to how JPA transactions persist objects)

to have reliable tests that actually pass and fail based on the code rather than flaws in the test

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitTest.kt
@@ -139,9 +139,12 @@ class BookVisitTest(@Autowired private val objectMapper: ObjectMapper) : Integra
 
     val visits = visitRepository.findAllByReference(reference)
     Assertions.assertThat(visits).hasSize(2)
-    Assertions.assertThat(visits[0].visitStatus).isEqualTo(CANCELLED)
-    Assertions.assertThat(visits[0].outcomeStatus).isEqualTo(OutcomeStatus.SUPERSEDED_CANCELLATION)
-    Assertions.assertThat(visits[1].visitStatus).isEqualTo(BOOKED)
+    val bookedEntity = visits.single { it.id == bookedVisit.id }
+    val reservedEntity = visits.single { it.id == reservedVisit.id }
+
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitTest.kt
```diff
@@ -139,9 +139,12 @@ class BookVisitTest(@Autowired private val objectMapper: ObjectMapper) : Integra
...
-    Assertions.assertThat(visits[1].visitStatus).isEqualTo(BOOKED)
+    val bookedEntity = visits.single { it.id == bookedVisit.id }
+    val reservedEntity = visits.single { it.id == reservedVisit.id }
+
+    Assertions.assertThat(bookedEntity.visitStatus).isEqualTo(CANCELLED)
+    Assertions.assertThat(bookedEntity.outcomeStatus).isEqualTo(OutcomeStatus.SUPERSEDED_CANCELLATION)
+    Assertions.assertThat(reservedEntity.visitStatus).isEqualTo(BOOKED)
 
     // And
     val visit = objectMapper.readValue(returnResult.responseBody, VisitDto::class.java)
```
AledWEvans, line 150: update the above test code, because I notice a seemingly random failure of the test, the order from DB (visits list) was random,... the new code is a better way of testing it as it ensures it uses the correct PK id for each row!

dhirajshettymoj: Agreed. Should we also update the findAllByReference to add an orderBy?

dhirajshettymoj (approved):
Looks good. So basically we are setting the session template weekday based on today's weekday right? Good approach. The tests should run on all weekdays now. Thanks.

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitTest.kt
AledWEvans: I did at one stage and then reverted it, as it's not used for anything else.. It should be moved into a TestRepo class TBO..

PR #198, opened 2022-11-02
VB-1352 - Revert some changes made in VB-1296 

dhirajshettymoj:

Revert some changes made in VB-1296 as multiple visits are being returned when there are more than 1 visitors for a visit. Reintroduce VisitFilter and VisitSpecification files and remove native query added in VisitRepository.

Fixes issue where multiple visits are being returned if there are more than 1 visitor added to a visit.


tpmcgowan (approved):
LGTM - have tested and it fixes the issue

PR #201, opened 2022-11-07
VB-1403 GET /visits - prisonId shouldn't be a required parameter

AledWEvans:

it changes prison id to not required, however if a prison Id not present then a prisoner id must be, this is to prevent the hole DB from being downloaded 

To allow the consumer of the API to download all visits for a prisoners


dhirajshettymoj (approved):
Looks good to me.

tpmcgowan (approved):
LGTM

PR #203, opened 2022-11-08
Vb 1387 return visit times dependent on prisoner location

dhirajshettymoj:

Code changes to incorporate wing based visit sessions.

Allows visit bookings in prisons which do wing based visits (e.g. Bristol).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PrisonerDetailDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PrisonerDetailDto.kt
@@ -4,7 +4,9 @@ import io.swagger.v3.oas.annotations.media.Schema
 
 @Schema(description = "Prisoner details - only offenderNo and location needed")
 open class PrisonerDetailDto(
+  @Schema(description = "The prisoner's unique offender number, naps to prisonerId on VSIP", example = "A0000AA", required = true)
   val offenderNo: String,
 
+  @Schema(description = "Name of the location where the prisoner resides (if in prison)", example = "WRI-B-3-018")
   val internalLocation: String
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PrisonerDetailDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PrisonerDetailDto.kt
@@ -1,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Prisoner details - only offenderNo and location needed")
+open class PrisonerDetailDto(
+  val offenderNo: String,
+
+  val internalLocation: String
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PrisonerDetailDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PrisonerDetailDto.kt
@@ -8,5 +8,5 @@ open class PrisonerDetailDto(
   val offenderNo: String,
 
   @Schema(description = "Name of the location where the prisoner resides (if in prison)", example = "WRI-B-3-018")
-  val internalLocation: String
+  val internalLocation: String?
 )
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PrisonerDetailDto.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Prisoner details - only offenderNo and location needed")
+open class PrisonerDetailDto(
+  val offenderNo: String,
```
AledWEvans, line 7: I think we call this prisoner id in VSiP, could use annotation to match it up with prisoner api , ur call

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -58,13 +60,26 @@ class SessionService(
     val today = LocalDate.now()
     val requestedBookableStartDate = today.plusDays(noticeDaysMin ?: policyNoticeDaysMin)
     val requestedBookableEndDate = today.plusDays(noticeDaysMax ?: policyNoticeDaysMax)
+    var prisonerWing: String?
```
AledWEvans, line 63: get rid of this and put the this down on line 74 as follows : 

`val prisonerWing = getPrisonerWing(it)`

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -58,13 +60,26 @@ class SessionService(
...
     )
 
+    // filter the sessionTemplates to use the ones that are eligible for all wings and sessionTemplates for that wing
+    prisonerId?.also {
+      prisonerWing = getPrisonerWing(it)
+
+      // return all sessions where prison wings list is empty OR
+      // sessions where wings contain the prisoners wing
+      sessionTemplates = sessionTemplates.filter { sessionTemplate ->
+        isSessionEligibleForAllWings(sessionTemplate) || (prisonerWing != null && isSessionEligibleForWing(sessionTemplate, prisonerWing))
```
AledWEvans, line 79: not sure about this,... is it an OR? I think I'm trying to work this out while I type.. if a wing has a certain session  template for it should it ignore other session templates for the same prison that does not have a wing configuration?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/utils/PrisonerWingCalculator.kt
```diff
@@ -0,0 +1,46 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.utils
+
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.PrisonerDetailDto
+
+/**
+ * this is only a temporary solution till we can get it from an API
+ */
+class PrisonerWingCalculator {
```
AledWEvans, line 8: It's not a Calculator it's a Parser,....
In fact I would move all this into a service.PrisionApiService
have the getPrisonerWing in this service.
Then just parse the wing in the service you could leave it as an internal class

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/utils/PrisonerWingCalculator.kt
```diff
@@ -0,0 +1,46 @@
...
+  private enum class WingElementByPrison(
```
AledWEvans, line 9: PrisonWingFormat might be a better name

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -236,4 +251,31 @@ class SessionService(
 
   private fun isDateWithinRange(sessionDate: LocalDate, startDate: LocalDate, endDate: LocalDate? = null) =
     sessionDate >= startDate && (endDate == null || sessionDate <= endDate)
+
+  private fun getPrisonerWing(prisonerId: String): String? {
```
AledWEvans, line 255: move into a PrisionApiService

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -58,13 +60,26 @@ class SessionService(
...
+    // get all the sessions for that prison - wing and non wing based
+    var sessionTemplates = sessionTemplateRepository.findValidSessionTemplatesByPrisonId(
       prisonId,
       requestedBookableStartDate,
       requestedBookableEndDate
     )
 
+    // filter the sessionTemplates to use the ones that are eligible for all wings and sessionTemplates for that wing
+    prisonerId?.also {
+      prisonerWing = getPrisonerWing(it)
```
AledWEvans, line 74: this should be 
`prisionApiService.getPrisonerWing(it)`

AledWEvans (changes requested):
Some changes are required! Also @rj-adams may have some changes for the Wing approach, because we use levels instead I would not merge this code until those issues are clear.

PR #204, opened 2022-11-09
VB-1413 add ithc IP addresses to allowlist 

dhirajshettymoj:

Add ithc IP addresses to allowlist to allow penetration testing to be carried out on preprod.

Adding the IP addresses would allow the health checks on API to go ahead.


dhirajshettymoj:
Will wait on confirmation about the third IP address before doing a merge.

tpmcgowan:
Won't this add these IPs to all environments, as it's in the default `values.yaml`? I _think_ you might have to duplicate the default `allowlist` with the new IPs added in `values-preprod.yaml` for this to have an effect on preprod only.

dhirajshettymoj:
don't think this is needed any more as they can access the API without adding these IPs. closing the PR.

PR #205, opened 2022-11-09
VB-1411 Turn off domain events on staging

AledWEvans:

turns off domain event's in staging

to prevent systems down stream from getting confused,... e.g. same visit references...


dhirajshettymoj (approved):
Looks good

PR #206, opened 2022-11-10
VB-1417 Remove Wing base structure from DB and DTO's to prepare for L…

AledWEvans:
…evel based structure

removes code that is not going to be used

Clean up code base


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/SessionTemplateDto.kt
```diff
@@ -33,9 +33,7 @@ data class SessionTemplateDto(
   @Schema(description = "open capacity", example = "50", required = true)
   val openCapacity: Int,
   @Schema(description = "day of week fpr visit", example = "MONDAY", required = false)
-  val dayOfWeek: DayOfWeek?,
-  @Schema(description = "List of prison wings associated with the session template, if list is empty then all wings are acceptable!", required = false)
-  val prisonWings: List<SessionPrisonWingDto> = mutableListOf()
+  val dayOfWeek: DayOfWeek?
```
dhirajshettymoj, line 36: sorry not to do with the current change, just noticed this. Should dayOfWeek here be not null as it cannot be null on DB.

dhirajshettymoj (approved):
This change is good to go. Only 1 thing I noticed - not related to this change though. Should dayOfWeek on the SessionTemplateDTO be not null as it cannot be null on DB.

AledWEvans:
Your correct but not part of this review..

PR #208, opened 2022-11-11
Vb 1358 fix nested tests

dhirajshettymoj:

Moves nested classes out of CancelVisitTest as these were not running properly.

 Fixes to tests to fix failures


AledWEvans (approved):
LGTM

PR #210, opened 2022-11-14
Feature/introduce PrisonApiService

dhirajshettymoj:

Adds a new class - PrisonApiService to make calls to Prison API.

Code improvements


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/PrisonApiService.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  fun getOffenderNonAssociationList(prisonerId: String): List<OffenderNonAssociationDetailDto> {
+    try {
+      val offenderNonAssociationList = prisonApiClient.getOffenderNonAssociation(prisonerId)?.nonAssociations ?: emptyList()
+      SessionService.LOG.debug("sessionHasNonAssociation prisonerId : $prisonerId has ${offenderNonAssociationList.size} non associations!")
+      return offenderNonAssociationList
+    } catch (e: WebClientResponseException) {
+      if (e.statusCode != HttpStatus.NOT_FOUND) {
+        LOG.error("Exception thrown from call to prison API : $e")
+        throw e
```
AledWEvans, line 28: why just NOT_FOUND, would it be useful to throw exception  in all exception cases!

AledWEvans (commented):
May be extend you exception handling for all exceptions other wise you may have a silent error occurring!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/PrisonApiService.kt
AledWEvans: I think this is because prison API uses http error codes incorrectly, when it cant find prison it return a 404. So after all this is needed. It should not do this it should return a 200 when it cant find something,,... no result status

AledWEvans (approved):
LGTM

PR #211, opened 2022-11-14
VB-1431 Add new prison codes to DB

AledWEvans:

add's permitted prison codes, if already thire it ignores. 

To add BRS prison to allow testing of new prison switcher


dhirajshettymoj (approved):
LGTM.

src/main/resources/db/migration/data/R__add_prison_codes.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO prison(code, active) SELECT 'HEI', true WHERE NOT EXISTS ( SELECT id FROM prison WHERE code = 'HEI');
+INSERT INTO prison(code, active) SELECT 'BRS', true WHERE NOT EXISTS ( SELECT id FROM prison WHERE code = 'BRS');
```
tpmcgowan, line 2: Code for Bristol is `BLI`

AledWEvans: will change asap

PR #214, opened 2022-11-15
VB-1418 Level / model for handling visit sessions constraint by the p…

AledWEvans:
…rison's housing structure (WIP)

Add's Level based structure to VSiP

To allow us to store sessions based on prisoner location


rj-adams:
I can see what you are doing with the "INCLUDING: and "EXCLUDING" and I'm guessing it is to reduce the number of rows in the DB. I did think about this when constructing the spreadsheet mode but thought it would make maintenance by a business user difficult.

I think it will be difficult to write a front end to manage the includes and excludes concepts and I think business users will struggle to understand how to configure it.

Maybe worth a discussion...

AledWEvans:
As discussed it would work but with extra complexity in the service layer (not too much), but is it required? I don't want to add if there is no need for it.. check with @Phoebe C or Simon N, happy to have one or the other as long as it's agreed. Will remove for now and start working on how to load these configurations.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PermittedSessionLocationDto.kt
```diff
@@ -0,0 +1,31 @@
...
+data class PermittedSessionLocationDto(
+
+  @Schema(description = "Permitted Session Location id", example = "123", required = true)
+  val id: Long,
+
+  @Schema(description = "Level one location code", example = "c", required = true)
+  var levelOneCode: String,
+
+  @Schema(description = "Level two location code", example = "1", required = false)
+  var levelTwoCode: String,
```
dhirajshettymoj, line 15: this and levelThreeCode and level4Code needs to be nullable

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
```diff
@@ -33,9 +32,6 @@ data class PermittedSessionLocation(
   @JoinColumn(name = "PRISON_ID", updatable = false, insertable = false)
   val prison: Prison,
 
-  @Column(name = "TYPE", nullable = false)
-  val type: PermittedType,
-
   @ManyToMany(mappedBy = "permittedSessionLocations")
   var sessionTemplates: MutableSet<SessionTemplate> = mutableSetOf(),
```
dhirajshettymoj, line 36: We could make this var private so as not to have a getter as we only want PermittedSessionLocations as part of SessionTemplate. But not sure if it will impact how JPA works.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
```diff
@@ -0,0 +1,56 @@
...
+  @JoinColumn(name = "PRISON_ID", updatable = false, insertable = false)
+  val prison: Prison,
+
+  @ManyToMany(mappedBy = "permittedSessionLocations")
+  var sessionTemplates: MutableSet<SessionTemplate> = mutableSetOf(),
+
+  @Column(name = "LEVEL_ONE_CODE", unique = false, nullable = false)
+  var levelOneCode: String,
+  @Column(name = "LEVEL_TWO_CODE", unique = false)
+  var levelTwoCode: String,
```
dhirajshettymoj, line 41: level 2,3 and 4 codes can be null

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
```diff
@@ -0,0 +1,56 @@
...
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToMany
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+import javax.persistence.TemporalType
+
+@Entity
+@Table(name = "PERMITTED_SESSION_LOCATION")
+data class PermittedSessionLocation(
```
dhirajshettymoj, line 21: This is something for later but  there is a warning that JPA entities should not be data class. So we will have to change all the entities on VSIP defined as data class.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/SessionTemplate.kt
```diff
@@ -33,6 +37,14 @@ data class SessionTemplate(
   @Column(nullable = false)
   val visitRoom: String,
 
+  @ManyToMany(fetch = FetchType.LAZY, cascade = [CascadeType.ALL])
```
dhirajshettymoj, line 40: Sorry we did discuss this yesterday but do you think OneToMany would be better here and have PermittedSessionLocation as a child table since we do not have the full list of Session locations and only add it adhoc based on the SessionTemplate. So it would be similar to a person adding the allowed location as text for a session template. Happy to discuss further.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/PermittedSessionLocationDto.kt
AledWEvans: Will change...

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
AledWEvans: agreed

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
AledWEvans: will change

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
AledWEvans: removed this now but rich reckons we should give it a go

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/SessionTemplate.kt
AledWEvans: It's a many to many, will save us a lot of time especially since we will have lot's of things to config,... it has it's down side but should be manage via font end,... will think about this tonight and have a chat tomorrow

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
AledWEvans: Should I put this back in then? I have the script done for this also... so if we do that it's done

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
```diff
@@ -0,0 +1,56 @@
...
+
+  @Column(name = "PRISON_ID", nullable = false)
+  val prisonId: Long,
+
+  @ManyToOne
+  @JoinColumn(name = "PRISON_ID", updatable = false, insertable = false)
+  val prison: Prison,
+
+  @ManyToMany(mappedBy = "permittedSessionLocations")
+  var sessionTemplates: MutableList<SessionTemplate> = mutableListOf(),
```
dhirajshettymoj, line 36: we could make this var private so that the getter is not exposed but am not sure if it impacts how JPA works.

src/main/resources/db/migration/V2_10__add_permitted_session_location_table.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE public.permitted_session_location (
+    id  serial NOT NULL PRIMARY KEY,
+    prison_id int8 NOT NULL,
+	level_one_code varchar(10) NOT NULL,
+	level_two_code varchar(10) NULL,
+    level_three_code varchar(10) NULL,
+	level_four_code varchar(10) NULL,
+    create_timestamp timestamp default current_timestamp,
+    modify_timestamp timestamp default current_timestamp,
+    CONSTRAINT fk_permitted_session_location_to_prison FOREIGN KEY (prison_id) REFERENCES public.prison(id)
```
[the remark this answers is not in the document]
AledWEvans: seem a bit much really... should be handled in code validation if anything, validate a first opportunity not last... so I would say do it on the DTO when we create the admin api.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/session/PermittedSessionLocation.kt
AledWEvans: Not a good idea with JPA objects as they are wrapped by proxy classes and when you create a new Session Template you would have to add it to an existing location object after it has been created by the JPA frame work so you would need to have access to sessionTemplates attribute to add after creation.

dhirajshettymoj (approved):

PR #216, opened 2022-11-16
Vb 1432 get prisoner information from prison api

dhirajshettymoj:

Adds call to Prison API - getPrisonerInformationById.

Preliminary work for getting sessions by level.


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -220,4 +221,21 @@ class SessionService(
 
   private fun isDateWithinRange(sessionDate: LocalDate, startDate: LocalDate, endDate: LocalDate? = null) =
     sessionDate >= startDate && (endDate == null || sessionDate <= endDate)
+
+  private fun getOffenderNonAssociationList(prisonerId: String): List<OffenderNonAssociationDetailDto> {
+    return try {
+      prisonApiService.getOffenderNonAssociationList(prisonerId)
+    } catch (e: Exception) {
```
dhirajshettymoj, line 228: catching the exception as I believe visit scheduler should still return sessions if PrisonAPI is down. What do you think?

AledWEvans: If it's down we don't want to allow non associated prisoners mixing,,... so I would not allow sessions to be created if prison API services is down...

AledWEvans (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
dhirajshettymoj: Thanks, I have reverted back the code now.

PR #219, opened 2022-11-22
Explicitly name spring boot application to prevent error

AledWEvans:
Have to add this to prevent the following errors in "Execution failed for task ':bootJarMainClassName'.
> Unable to find a single main class"

_Required._

_Required._

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
 }
 
 springBoot {
-  mainClass.value("uk.gov.justice.digital.hmpps.visitscheduler.VisitScheduler")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -6,6 +6,10 @@ plugins {
   idea
 }
 
+springBoot {
+  mainClass.value("uk.gov.justice.digital.hmpps.visitscheduler.VisitScheduler")
```
AledWEvans, line 10: should this be VisitSchedulerApplication or VisitSchedulerApplicationKt I have no idea but it does say mainClass,.. so I'm assuming it's VisitScheduler, only one way to find out

AledWEvans: found out it was the file not the class bizzar format

dhirajshettymoj (approved):
LGTM

PR #220, opened 2022-11-23
VB-1436 send more data to app insights to allow better reporting.

dhirajshettymoj:

Additional data sent to application insights.

To allow better reporting on app insights.


AledWEvans (approved):
LGTM

PR #222, opened 2022-11-23
Vb 1435 add sql scripts to configure bi weekly session template for hmp bristol

AledWEvans:

Creates SQL with bi weekly / alternate data also added group to link table

Because Bristol requires these details to run


AledWEvans:
The data needs to be updated when we know when we are going live!

src/main/resources/db/migration/V2_12__add_group_to_link_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE session_to_permitted_location ADD location_group VARCHAR(80);
```
dhirajshettymoj, line 1: whats this new column for and does it need any code changes?

AledWEvans: No real code changes was going to introduce it at a later stage when we were doing the admin implementation, but had a convo with @rj-adams yesterday and thought it might be easer to store this data now rather than later so I could just do an auto migration in the Fly script. But just dont want to approach the hole subject now as it needs proper a analysis.

dhirajshettymoj (approved):
LGTM

PR #223, opened 2022-11-24
Improvement to sql generator to insert in session template tables in …

AledWEvans:
…correct order.

Tech debt, makes sure the Session templates are inserted into the table in a certain order

To make DB natural order logical from a reading point of view!


rj-adams:
OK but the database won't guarantee that they come out in the same order that they go in. You would need an "order by" cause on their way out to guarantee this.

dhirajshettymoj (approved):
LGTM

AledWEvans:
I know @rj-adams it's just cosmetic really... so it's reads better

PR #225, opened 2022-11-25
Improvement to sql generator, to use freemarker template

AledWEvans:

removes SQL in code and put's it into a free marker template

To separate code from output to follow POC

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,15 +24,16 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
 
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -24,15 +24,16 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
 
-  implementation("org.springdoc:springdoc-openapi-webmvc-core:1.6.12")
```
AledWEvans, line 27: looks like these needed updating to newer versions

dhirajshettymoj (approved):
Nice implementation. Reduces code.

PR #227, opened 2022-11-28
Vb 1453 get prisoner levels from new prison api endpoint

dhirajshettymoj:

Allows calling the new Prison API endpoint to get housing locations.
To incorporate new Prison API endpoint and introduce configuration to go back to old Prison API call in case there's an issue with the new endpoint.


rj-adams:
this should not go back to the "old" API - the old API was not suitable for our purpose.

If the "new" API fails, raise an exception and deal with in exactly the same way as a failure from any endpoint

AledWEvans:
@dhirajshettymoj to go back to the old API, seems like defensive programming which should be avoided. I am wondering why you are doing this, did the new API have errors? If it does it's not ready for release and a defect should be razed. For them to fix!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/PrisonApiService.kt
```diff
@@ -45,4 +51,24 @@ class PrisonApiService(
...
+        throw e
+      }
+    }
+
+    return null
+  }
+
+  fun getPrisonerDetails(prisonerId: String): PrisonerDetailsDto? {
+    return if (prisonApiLevelsEndpoint == PrisonApiGetLevelsEndpoint.FULL_STATUS)
+      getPrisonerFullStatus(prisonerId)
```
AledWEvans, line 70: Not sure we should be doing this, once new api is working remove old methods..

dhirajshettymoj:
@rj-adams @AledWEvans I was wondering just in case. the new endpoint has not been tested enough. But I guess now that they have gone Live it should have been tested. I will remove the code around the old Prison API endpoint call and resubmit the PR.

AledWEvans:
Well if there system is not ready in the Agile sense we use the old approach until it is, but if we have no reason or evidence to suggest it is not work, we simple move the pre PR code into preprod and then do testing on this code if there are issues we rase bugs accordingly.. So my suggestion is move the branch main in to preprod (ask team) , remove the dual code in this review and then test!

dhirajshettymoj:
@AledWEvans @rj-adams I have now updated the code to only use the new Prison API endpoint.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/client/PrisonApiClient.kt
```diff
@@ -26,11 +26,11 @@ class PrisonApiClient(
...
   }
 
-  fun getPrisonerDetails(offenderNo: String): PrisonerDetailDto? {
+  fun getPrisonerHousingLocation(offenderNo: String): PrisonerHousingLocationsDto? {
     return webClient.get()
-      .uri("/api/prisoners/$offenderNo/full-status")
+      .uri("/api/offenders/$offenderNo/housing-location")
       .retrieve()
-      .bodyToMono(prisonerDetail)
+      .bodyToMono(prisonerHousingLocationsDto)
```
AledWEvans, line 33: The use of bodyToMono can cause problems I think, if nothing comes back from the API you will get an empty object,... is that what we need. But since we are already using this in other calls maybe we will wait to see if we get any problems and fix as and when.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/prison/api/PrisonerHousingLocationsDto.kt
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  fun getHousingLevelByLevelNumber(housingLevel: Int): PrisonerHousingLevelDto? {
+    return levels.stream().filter { level -> level.level == housingLevel }.findFirst().orElse(null)
+  }
+
+  fun getLevels(): Map<PrisonerHousingLevels, String?> {
+    return levelsMap.toMap()
+  }
+}
```
AledWEvans, line 24: Move functionality out of this DTO, DTO should not have functionality, put functionality / levelsMap in service!

AledWEvans (changes requested):
Have maid comment's expect updates, will continue review once changes have been applied

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/prison/api/PrisonerHousingLocationsDto.kt
dhirajshettymoj: Updated and moved into the PrisonAPIService.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/client/PrisonApiClient.kt
dhirajshettymoj: Yes, this would be something we will need to look at at a later stage.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/PrisonApiService.kt
dhirajshettymoj: Updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/PrisonApiService.kt
```diff
@@ -33,16 +34,31 @@ class PrisonApiService(
...
-        LOG.error("Exception thrown on prison API call - /api/prisoners/$prisonerId/full-status", e)
+        LOG.error("Exception thrown on prison API call - /api/offenders/$prisonerId/housing-location", e)
         throw e
       }
     }
 
     return null
   }
+
+  fun getLevelsMapForPrisoner(prisonerHousingLocationsDto: PrisonerHousingLocationsDto): Map<PrisonerHousingLevels, String?> {
```
AledWEvans, line 50: nice

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/mock/PrisonApiMockServer.kt
```diff
@@ -65,38 +71,79 @@ class PrisonApiMockServer : WireMockServer(8092) {
...
-              """.trimIndent()
+              getHousingLocationJsonResponse(levels)
             )
         )
     )
   }
+
+  private fun getHousingLocationJsonResponse(levels: Map<Int, String?>): String {
+
+    val jsonResponse = StringBuilder("")
```
AledWEvans, line 92: could use object mapper for this, but this is fine

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/mock/PrisonApiMockServer.kt
```diff
@@ -65,38 +71,79 @@ class PrisonApiMockServer : WireMockServer(8092) {
...
+        }
+        levelsText.append(getLevelText(level, code))
+      }
+    }
+
+    return levelsText.toString()
+  }
+
+  private fun getLevelText(level: Int, code: String): String {
+    return """
```
AledWEvans, line 126: would be easier to use object mapper but this is fine, it's only test code

AledWEvans (approved):
LGTM, a lot of work here, nice work

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/mock/PrisonApiMockServer.kt
dhirajshettymoj: Used it initially. But it was failing on odd occasions. Will add that as tech debt ticket.

PR #229, opened 2022-11-30
VB-1471 OWASP Dependency-Check Analyzer, second try dependancy was no…

AledWEvans:
…t in uk.gov.justice.hmpps.gradle-spring-boot project,.. silly me

updated postgresql-42.5.0.jar to postgresql-42.5.1.jar

Makes sure we use latest versions of code for security reasons


dhirajshettymoj (approved):
LGTM

PR #230, opened 2022-11-30
VB-1479 Biweekly session for Bristol - Sunday uses the opposite week

AledWEvans:

Fixes a defect that skipped the last day of week on bi-weekly

fixes defect


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/session/GetSessionsTest.kt
```diff
@@ -51,6 +54,42 @@ class GetSessionsTest(@Autowired private val objectMapper: ObjectMapper) : Integ
     assertSession(visitSessionResults[0], nextAllowedDay, sessionTemplate)
   }
 
+  @Test
+  fun `bi weekly schedule - test week (Sunday) change boundary`() {
+    // If this test is run on a weekend it will FAIL...
```
dhirajshettymoj, line 59: This could be an issue if we decide to do an urgent release on the weekend. Could we add a Conditional annotation to stop it running on weekends if it will fial on weekends?

https://www.baeldung.com/junit-5-conditional-test-execution

AledWEvans: could put a conditional check on it!

AledWEvans: improved logic and added new tests, also added a switch to make sure it works at week end, tested by changing system date

dhirajshettymoj (approved):

PR #231, opened 2022-12-01
VB-1451 Extend service to allow 'cancel' and 'booked' filtered visits

AledWEvans:

adds functionality to search on more than one status

to apply requirements in stories and from user research

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
@@ -32,6 +32,7 @@ import java.time.LocalDateTime
 import javax.validation.Valid
 
 const val VISIT_CONTROLLER_PATH: String = "/visits"
+const val VISIT_CONTROLLER_SEARCH_PATH: String = "$VISIT_CONTROLLER_PATH/search"
 const val VISIT_RESERVE_SLOT: String = "$VISIT_CONTROLLER_PATH/slot/reserve"
 const val VISIT_RESERVED_SLOT_CHANGE: String = "$VISIT_CONTROLLER_PATH/{applicationReference}/slot/change"
 const val VISIT_CHANGE: String = "$VISIT_CONTROLLER_PATH/{reference}/change"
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitControllerLegacy.kt
```diff
@@ -73,40 +73,40 @@ class VisitControllerLegacy(
     ) prisonerId: String?,
     @RequestParam(value = "prisonId", required = false)
     @Parameter(
```
AledWEvans, line 75: @tpmcgowan this is the old endpoint you were using, the new one or not so new was being hidden by swagger because it had a defect in it (swagger).

tpmcgowan:
As discussed, @AledWEvans, it would make sense for `visitStatus` (on old and new endpoint) to be a required field - as we're always calling it with this anyway.

AledWEvans:
> As discussed, @AledWEvans, it would make sense for `visitStatus` (on old and new endpoint) to be a required field - as we're always calling it with this anyway.

Sure done and uploaded

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/VisitFilter.kt
```diff
@@ -5,12 +5,8 @@ import java.time.LocalDateTime
 data class VisitFilter(
   val prisonerId: String? = null,
   val prisonCode: String? = null,
```
dhirajshettymoj, line 7: should we change this to a non optional field and startdatetime and enddatetime?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
```diff
@@ -300,31 +301,31 @@ class VisitController(
...
       example = "12322"
-    ) nomisPersonId: Long?,
-    @RequestParam(value = "visitStatus", required = false)
+    ) visitorId: Long?,
+    @RequestParam(value = "visitStatus", required = true)
     @Parameter(
       description = "Filter results by visit status",
       example = "BOOKED"
-    ) visitStatus: VisitStatus?,
+    ) visitStatusList: List<VisitStatus>,
```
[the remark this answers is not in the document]
AledWEvans: I did think about this! not sure really, certainly no need for CHANGING!

PR #232, opened 2022-12-01
VB-1477 Christmas Timetable for HMP Hewell

AledWEvans:

Adds Xmas time table for Hewell

To configure Hewell prison for the Xmas break because schedules have changed for this period


src/main/resources/session-template-data/template.ftl
```diff
@@ -67,7 +67,7 @@ UPDATE tmp_permitted_session_location SET prison_id = prison.id FROM prison WHER
 INSERT INTO permitted_session_location(id,prison_id,level_one_code,level_two_code,level_three_code,level_four_code)
 SELECT id,prison_id,level_one_code,level_two_code,level_three_code,level_four_code FROM tmp_permitted_session_location order by id;
 
-ALTER SEQUENCE permitted_session_location_id_seq RESTART WITH 97;
+ALTER SEQUENCE permitted_session_location_id_seq RESTART WITH ${permitted_session_location_index};
```
AledWEvans, line 70: now automated

rj-adams: do you need this line at all - why not just let the sequences run their course and keep incrementing - they are just unique codes, they don't need to start from 1 or be contiguous.

AledWEvans: Yes we do, when the id is inserted and not auto generated it stays at 1 other wise causes all sorts of issues... must be a postgresql thing.

dhirajshettymoj (approved):
LGTM

PR #233, opened 2022-12-01
Vb 1480 visits are shown for the wrong prison in update booking

dhirajshettymoj:

Does not return sessions if the prisonId matched does not match prisonerId.

To fix a bug around wrong sessions being returned for a prisoner


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/mock/PrisonApiMockServer.kt
```diff
@@ -81,69 +85,49 @@ class PrisonApiMockServer : WireMockServer(8092) {
             .withHeader("Content-Type", MediaType.APPLICATION_JSON_VALUE)
             .withStatus(200)
             .withBody(
-              getHousingLocationJsonResponse(levels)
+              getJsonString(housingLocationsDto)
```
dhirajshettymoj, line 88: Moved this to ObjectMapper.toJsonString() as suggested in the last review.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/exception/ExpiredVisitAmendException.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.exception
```
dhirajshettymoj, line 1: Existing exceptions - moved them out into a separate package.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -53,30 +54,42 @@ class SessionService(
     val requestedBookableStartDate = today.plusDays(noticeDaysMin ?: policyNoticeDaysMin)
     val requestedBookableEndDate = today.plusDays(noticeDaysMax ?: policyNoticeDaysMax)
 
+    // ensure the prisoner - if supplied belongs to the same prison as supplied prisonCode
+    prisonerId?.let {
+      prisonerValidationService.validatePrisonerIsFromPrison(prisonerId, prisonCode)
+    }
```
AledWEvans, line 60: Not really for this review but why is prisonerId optional...

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
```diff
@@ -53,30 +54,42 @@ class SessionService(
...
       requestedBookableEndDate
     )
-
     sessionTemplates = filterSessionsTemplatesForLocation(sessionTemplates, prisonerId)
 
     var sessions = sessionTemplates.map {
       buildVisitSessionsUsingTemplate(it, requestedBookableStartDate, requestedBookableEndDate)
     }.flatten()
 
-    if (!prisonerId.isNullOrBlank()) {
```
AledWEvans, line 68: I would revert this back, because the original code needs improving and the code is defensive

AledWEvans (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionService.kt
dhirajshettymoj: Yes, need to pick this up as tech debt. We could also remove prisonId parameter and only use prisonerId and get the prisoner's prison ID from a Prison API call so that we always get the sessions for the right prison if the front end does not use only prisonId option.

AledWEvans (approved):
LGTM

PR #235, opened 2022-12-05
VB-1487 Update visits for xmas script

AledWEvans:

Added a script to alter times of Xmas period,..

To fix timetable to xmas time table


dhirajshettymoj (approved):
Approved. Do we plan to delete these scripts later from the codebase?

PR #236, opened 2022-12-05
VB-1488 Christmas Timetable for HMP Bristol

AledWEvans:

Adds configuration for Xmas time table for Bristol

Because Bristol has a different time table over Xmas period


dhirajshettymoj (approved):
LGTM

PR #243, opened 2022-12-13
Vb 1486 update entities to ensure they are not data classes

dhirajshettymoj:

Change to update entities from data classes to open classes.
Performance improvements and based on JPA recommendations


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/Visit.kt
```diff
@@ -99,12 +99,10 @@ data class Visit(
 
   @Column
   var reference = _reference
-    private set
```
dhirajshettymoj, line 102: had to remove this private set as open classes cannot have private set methods it seems.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/Prison.kt
```diff
@@ -35,4 +35,20 @@ data class Prison(
   @Temporal(TemporalType.TIMESTAMP)
   @Column
   val modifyTimestamp: LocalDateTime? = null
-)
+) {
+  override fun equals(other: Any?): Boolean {
```
dhirajshettymoj, line 39: @AledWEvans - Haven't added the parent class for entities as you suggested yesterday. Could we pick that up as part of  different code improvements change?

AledWEvans: sure

PR #244, opened 2022-12-13
VB-1486 - change all entities from data classes to open classes

dhirajshettymoj:

change all entities from data classes to open classes as pr JPA recommendations.

performance improvements and to follow recommendations.


AledWEvans (approved):
LGTM

PR #249, opened 2022-12-21
VB-1537 Some visits have turned up with a status of BOOKED and a outcome status of SUPERSEDED CANCELLATION

AledWEvans:
Some visits have turned up with a status of BOOKED and a outcome status of SUPERSEDED CANCELLATION

Fixes bug found in https://dsdmoj.atlassian.net/browse/VB-1537

To prevent DB data from being inaccurate...

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -135,7 +135,7 @@ class VisitService(
   }
 
   fun changeVisitSlot(applicationReference: String, changeVisitSlotRequestDto: ChangeVisitSlotRequestDto): VisitDto {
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -227,7 +227,7 @@ class VisitService(
 
   fun bookVisit(applicationReference: String): VisitDto {
 
-    val visitToBook = visitRepository.findByApplicationReference(applicationReference) ?: throw VisitNotFoundException("Could not find reserved visit applicationReference:$applicationReference not found")
+    val visitToBook = visitRepository.findApplication(applicationReference) ?: throw VisitNotFoundException("Application (reference $applicationReference) not found")
```
dhirajshettymoj, line 230: Is the "Application (reference $applicationReference) not found" incorrect in this scenario as the visit might have been BOOKED? Should we instead check the visit status? Or set the message to not found or already booked?

dhirajshettymoj (approved):

PR #252, opened 2023-01-09
VB-1563 Update HMP Bristol visits for 31/12/22

AledWEvans:

Changes some past visits to correct date and times

to fix up errors from migrated and created bookings..


dhirajshettymoj (approved):
LGTM

PR #253, opened 2023-01-09
Vb 1563 update hmp bristol visits for 31/12/22

AledWEvans:

configures SQL generator to solve the problem in VB-1563

to fix the bug


src/main/resources/db.scripts.mvp/fix_up_past_visits_for_bristol_on_december_31st.sql
```diff
@@ -0,0 +1,11 @@
+-- VB-1598 View establishment schedule change from 9:30/10:30 to 15:30/16:30
```
AledWEvans, line 1: whats this doing here this should be part of develop

dhirajshettymoj (approved):

PR #254, opened 2023-01-11
VB-1658 Upcoming visits search: Booked and Cancelled visits with same reference

AledWEvans:
Upcoming visits search: Booked and Cancelled visits with same reference

Prevents visits with outcome status of SUPERSEDED_CANCELLATION from being returned

to fix the defect.


dhirajshettymoj (approved):
LGTM

PR #257, opened 2023-01-16
Vb 1494 add validation to restrict number of visitors

dhirajshettymoj:
Setting an upper limit (10) to the number of visitors allowed on a visit.

This validation will ensure that any invalid requests with loads of visitors (more than 10) will be validated and hence not allowed


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/validators/VisitorCountValidator.kt
```diff
@@ -0,0 +1,37 @@
...
+      this.maximumVisitorsAllowed = constraintAnnotation.maximumVisitorsAllowed
+    }
+  }
+
+  override fun isValid(childList: Set<VisitorDto>?, context: ConstraintValidatorContext?): Boolean {
+    // Cannot have more than max visitors allowed
+    val total = (childList?.size ?: 0).toUShort()
+    return (total <= maximumVisitorsAllowed)
+  }
+}
```
AledWEvans, line 37: I would do the following
```

package uk.gov.justice.digital.hmpps.visitscheduler.controller.validators

import uk.gov.justice.digital.hmpps.visitscheduler.dto.VisitorDto
import javax.validation.Constraint
import javax.validation.ConstraintValidator
import javax.validation.ConstraintValidatorContext
import javax.validation.Payload
import kotlin.reflect.KClass

@Target(AnnotationTarget.FIELD)
@MustBeDocumented
@Constraint(validatedBy = [VisitorCountValidator::class])
annotation class VisitorCountValidation(
  val message: String = "{javax.validation.constraints.visitor.count.message}",
  val groups: Array<KClass<*>> = [],
  val payload: Array<KClass<out Payload>> = [],
  val maximumVisitorsAllowed: Int = 10,
)

class VisitorCountValidator : ConstraintValidator<VisitorCountValidation, Set<VisitorDto>> {

  private var maximumVisitorsAllowed: Int = 0

  override fun initialize(constraintAnnotation: VisitorCountValidation) {
      this.maximumVisitorsAllowed = constraintAnnotation.maximumVisitorsAllowed
  }

  override fun isValid(visitors: Set<VisitorDto>?, context: ConstraintValidatorContext?): Boolean {
    // Cannot have more than max visitors allowed
    val totalVisitors = visitors?.size ?: 0
    return (totalVisitors <= maximumVisitorsAllowed)
  }
}


```

Just to make it simple and straight forward

AledWEvans (changes requested):
some possible minor changes,

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/validators/VisitorCountValidator.kt
dhirajshettymoj: Thanks, updated based on your comments.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/validators/VisitorCountValidator.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlin.reflect.KClass
+
+@Target(AnnotationTarget.FIELD)
+@MustBeDocumented
+@Constraint(validatedBy = [VisitorCountValidator::class])
+annotation class VisitorCountValidation(
+  val message: String = "{javax.validation.constraints.visitor.count.message}",
+  val groups: Array<KClass<*>> = [],
+  val payload: Array<KClass<out Payload>> = [],
+  val maximumVisitorsAllowed: Short = 10
```
AledWEvans, line 17: Short, good call

AledWEvans (approved):
LGTM

PR #264, opened 2023-01-25
Vb 1738 tech debt remove room name check on book counts logic

AledWEvans:

Removes unused logic

To prevent room logic affecting booking counts


dhirajshettymoj (approved):
LGTM

PR #270, opened 2023-02-01
VB-1763 - script to update visits for Cardiff, Preston and Wandsworth

dhirajshettymoj:
VB-1763 - script to update visits for Cardiff, Preston and Wandsworth to ensure visit times to match the timetables provided by each prison.

Updates visits migrated for the 3 new prisons to be onboarded to update the visit timings with the data in the session templates.

Keep visit data in sync with session templates to allow these visits to be updated.


dhirajshettymoj:
WIP - Adding a PR for initial review. Will do some more checks on Live before merging it.

AledWEvans:
Looks good to me, I would have used a join instead of a subquery but as this is practically throw away code it does not matter... looks like sound SQL

dhirajshettymoj:
@AledWEvans - Removed the WIP from PR. Could you please approve.

AledWEvans (approved):
LGTM, no changes from yesterday... I assume no changes for the prod migration

dhirajshettymoj:
THis was tested on preprod and looking at the data we can run these scripts in Prod now.

PR #271, opened 2023-02-02
Vb 1766 delete cardiff video link visits from vsip

dhirajshettymoj:
Post data migration it was noticed that Video Link visits created on NOMIS have been migrated as part of VSIP migration. It was decided to delete these records as they need not be on VSIP.

Data sanity.


rj-adams (commented):
Why include the 30 minute restriction - I would have thought that we want to delete all Video link visits regardless of length

AledWEvans (approved):
LGTM, very nice sql ;-)

dhirajshettymoj:
> Why include the 30 minute restriction - I would have thought that we want to delete all Video link visits regardless of length

just did a count and both return 47 records. Will update the script to remove the 30 minute constraint.

rj-adams (dismissed):
lgtm

AledWEvans (approved):
LGTM

PR #272, opened 2023-02-02
VB-1766 - delete 5 future dated visits.

dhirajshettymoj:

Deletes 5 specific future dated visits.

Prison specific request.


AledWEvans (commented):
LGTM as long as references are good should work just fine!

AledWEvans (approved):
Approved

PR #278, opened 2023-02-27
VB-1878 Server side to return Enhanced in DTO

AledWEvans:

Returns Enhanced in prison schedule

to allow the UI to display the correct data to the user


dhirajshettymoj (approved):
LGTM

PR #284, opened 2023-03-09
VB-1981 - updates to allow updated visits to be cancelled.

dhirajshettymoj:

Updates code to allow cancellation of cancelled visits.

Fix for VB-1981.


AledWEvans (approved):
LGTM

PR #290, opened 2023-03-13
Vb 1921 user information from request header and save on database

dhirajshettymoj:

Gets the user name information from request header instead of token as the orchestration service will connect to visit-scheduler via client token.

To reflect changes to the way the orchestration service will connect to visit-scheduler.


rj-adams:
I'm wondering if we need this extra header:
If the client ID contains the “sub” to the user, then is this sufficient? is it necessary to also pass the username as a separate header?

or do I misunderstand this

dhirajshettymoj:
Code updated to get actionedBy from request body for reserve and cancel visit. Also merged from develop branch.

build.gradle.kts
```diff
@@ -59,6 +59,7 @@ dependencies {
   testImplementation("org.springframework.security:spring-security-test")
   testImplementation("org.testcontainers:postgresql:1.17.6")
   testImplementation("org.testcontainers:localstack:1.17.6")
+  testImplementation("com.amazonaws:aws-java-sdk-s3:1.12.13")
```
AledWEvans, line 62: odd this is, I don't have this on mine and everything works, Has this got something to do with the docker/local host stuff you did?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitControllerLegacy.kt
```diff
@@ -78,8 +79,8 @@ class VisitControllerLegacy(
     @PathVariable
     reference: String,
     @RequestBody @Valid
-    cancelOutcome: OutcomeDto,
+    cancelVisitDto: CancelVisitDto,
   ): VisitDto {
-    return visitService.cancelVisit(reference.trim(), cancelOutcome)
+    return visitService.cancelVisit(reference.trim(), cancelVisitDto)
```
AledWEvans, line 84: do we want to put it on legacy, as syscon may be still using it, I think they might have started using the new rest end point probably worth checking! I've asked @lbennett-moj  about it!

AledWEvans (approved):
LGTM  but would check with syscon and perhaps move old way into legacy if no longer using it.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitControllerLegacy.kt
dhirajshettymoj: thanks.. Yes I will speak to syscon

PR #291, opened 2023-03-15
Vb 2002 admin api tasks server side dummy request response m

AledWEvans:

Adds element to UI for prisoner Categories, but dummies data

to allow other services to update there code without waiting for server side dev


dhirajshettymoj (approved):
LGTM

PR #301, opened 2023-03-21
VB-2021 Update SQL generator - Category

AledWEvans:

Update to SQL generator to allow for Category addition 

to implement the given storey


src/main/resources/session-template-data/template.ftl
```diff
@@ -1,121 +1,161 @@
...
+
+        -- Prisoner category
+
+        -- Create temporary group table
+            CREATE TABLE tmp_session_prisoner_category (
+                id                  serial          NOT NULL PRIMARY KEY,
+                code                varchar(100)    NOT NULL
+            );
+
+        <#if categories?has_content>
```
dhirajshettymoj, line 129: just for clarity - will this create duplicate codes? Should code be unique on the table so that its constant across prisons?

AledWEvans: No duplicates the code puts the categories in a set in the SQL generator!

rj-adams (approved):
Categories discussed and all fine

PR #305, opened 2023-03-24
Make actioned by feild optional

AledWEvans:

createdBy is now optional also added tests.

Sync process will be temporarily using old api without the created by details.


lbennett-moj (approved):
LGTM

PR #307, opened 2023-03-24
Merge develop into main

AledWEvans:

small change in develop that needed to be merged into main

so we can release with features covered in sprint

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,6 @@ ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 
 RUN apt-get update && \
     apt-get -y upgrade && \
-    apt-get install -y curl && \
     rm -rf /var/lib/apt/lists/*
 
 ENV TZ=Europe/London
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -20,7 +20,6 @@ ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 
 RUN apt-get update && \
     apt-get -y upgrade && \
-    apt-get install -y curl && \
```
dhirajshettymoj, line 23: should this come from main branch?

applicationinsights.dev.json
```diff
@@ -28,22 +28,6 @@
             }
           ],
           "percentage": 10
-        },
```
dhirajshettymoj, line 31: same as above

dhirajshettymoj (approved):

Dockerfile
dhirajshettymoj: Based on discussion with Aled, this is to be removed as per his chat with Peter Philips.

applicationinsights.dev.json
dhirajshettymoj: Based on discussion with Aled, this is to be removed as per his chat with Peter Philips.

PR #312, opened 2023-03-27
VB-1926 Update orchestration service to acquire full username and ret…

AledWEvans:
…urn it to the UI for each entry - added new end point to return visit history

allows us to return the history for a booking

the intent is to allow users to see how did what and when

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
@@ -38,7 +38,7 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
   @Query(
     "SELECT * FROM visit " +
       "WHERE reference = :reference AND visit_status IN ('BOOKED','CANCELLED')  " +
-      "ORDER BY modify_timestamp",
+      "ORDER BY id",
     nativeQuery = true,
   )
   fun findAllByReference(reference: String): List<Visit>
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
@@ -36,6 +36,14 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
   fun findByReference(reference: String): Visit?
 
   @Query(
+    "SELECT * FROM visit " +
+      "WHERE reference = :reference AND visit_status IN ('BOOKED','CANCELLED')  " +
+      "ORDER BY modify_timestamp",
+    nativeQuery = true,
+  )
+  fun findAllByReference(reference: String): List<Visit>
+
+  @Query(
     "SELECT v FROM Visit v WHERE v.applicationReference = :applicationReference AND (v.visitStatus = 'CHANGING' OR v.visitStatus = 'RESERVED') ",
   )
   fun findApplication(applicationReference: String): Visit?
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -35,6 +35,14 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
   )
   fun findByReference(reference: String): Visit?
 
+  @Query(
+    "SELECT * FROM visit " +
+      "WHERE reference = :reference AND visit_status IN ('BOOKED','CANCELLED')  " +
+      "ORDER BY modify_timestamp",
```
dhirajshettymoj, line 41: should this be create_timestamp or will it not matter?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -35,6 +35,14 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
...
+    nativeQuery = true,
```
dhirajshettymoj, line 42: does this need to be a native query?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -35,6 +35,14 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
   )
   fun findByReference(reference: String): Visit?
 
+  @Query(
+    "SELECT * FROM visit " +
```
dhirajshettymoj, line 39: I think this query will also return the SUPERSEDED_CANCELLATION records. The query might need to be updated to excluded them.

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/VisitHistoryByReferenceTest.kt
```diff
@@ -0,0 +1,48 @@
...
+    visitEntityHelper.create(reference = reference, prisonerId = "FF0000AA", visitStatus = CHANGING, visitStart = visitTime)
+
+    // When
+    val responseSpec = callVisitHistoryByReference(webTestClient, reference, roleVisitSchedulerHttpHeaders)
+
+    // Then
+    responseSpec.expectStatus().isOk
+      .expectBody()
+      .jsonPath("$.length()").isEqualTo(2)
+      .jsonPath("$[0].reference").isEqualTo(reference)
```
dhirajshettymoj, line 45: should we also test for the status to ensure only BOOKED and CANCELLED are returned?  That will also help ensure that the records are returned in the right order.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
AledWEvans: SUPERSEDED_CANCELLATION is the outcome status

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
AledWEvans: it's easy this way

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/VisitHistoryByReferenceTest.kt
rj-adams: Also, one should never assume database records will be returned in a specific order unless that order is specified. Suggest putting an"order by" clause in the SQL or whatever JPA requires to make that happen

AledWEvans: There is an order on the SQL statment!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
dhirajshettymoj: As discussed, yes we need to show these visits.

dhirajshettymoj (approved):
LGTM

PR #313, opened 2023-03-29
Remove apr the 6th from hmp morton hall schedule

AledWEvans:

We need to remove the following days from HMP Morton Halls social visits schedule as they have Family visits for those days.

If there are any visits booked for the dates, they should be cancelled.


dhirajshettymoj (approved):
LGTM - just wondering if we need to add anything to Visit_Notes for the cancelled visits but that is not needed as such.

AledWEvans:
I thought about notes but just put in outcome status instead..

PR #316, opened 2023-03-30
Update validate to large resource class in config.yml

dhirajshettymoj:

Update validate to large resource class in config.yml as suggested by CircleCI

Circle CI suggestion for memory issues


AledWEvans (approved):
LGTM, worth a go

PR #322, opened 2023-04-05
VB-2041 Update to TransitionalLocations check

AledWEvans:

It makes sure it's check for temp location ignores case, also adds further tests

to make code stable


dhirajshettymoj (approved):
LGTM

PR #323, opened 2023-04-05
VB-2041 Removed code that was not needed

AledWEvans:

removed code that had no function

to clean up code base


dhirajshettymoj (approved):
LGTM

PR #325, opened 2023-04-06
VB-2096 - update the applicationinsights.json to reduce logging

dhirajshettymoj:
update the applicationinsights.json to reduce logging on PreProd and Prod as it is logging all trace logs from dependencies as well.

updates the applicationinsights.json to reduce logging

 Too much trace logs being logged in application insights especially for dependencies like JPA, Hibernate etc.

### DIFF
--- a/applicationinsights.json
+++ b/applicationinsights.json
@@ -28,22 +28,6 @@
             }
           ],
           "percentage": 10
-        },
-        {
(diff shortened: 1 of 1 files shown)


AledWEvans (approved):
LGTM

rj-adams:
Is this something which could/should be incorporated into the Template App?

PR #326, opened 2023-04-06
VB-2096 - Reduce dependency logging on prod and preprod

dhirajshettymoj:
removed the sampling as can still see the dependency logs o…n preprod.

Reduce dependency logging on prod and preprod

Reduce dependency logging as logs are being cluttered


AledWEvans (approved):
LGTM

PR #327, opened 2023-04-06
VB-2096 - update to logback-spring and applicationinsights.json

dhirajshettymoj:
update to logback-spring to remove logappender and updates to applicationinsights.json.

Updates to logback-spring to reduce logging

Reduce logging on preprod and prod

### DIFF
--- a/applicationinsights.json
+++ b/applicationinsights.json
@@ -13,6 +13,9 @@
   "selfDiagnostics": {
     "destination": "console"
   },
+  "sampling": {
+    "percentage": 100
+  },
   "preview": {
     "sampling": {
       "overrides": [
--- a/applicationinsights.json
+++ b/applicationinsights.json
@@ -6,12 +6,18 @@
     "service.version": "${BUILD_NUMBER}"
   },
   "instrumentation": {
+    "springScheduling": {
+      "enabled": false
+    },
     "logging": {
       "level": "DEBUG"
     }
   },
   "selfDiagnostics": {
     "destination": "console"
+  },
+  "sampling": {
+    "percentage": 100
   },
   "preview": {
     "sampling": {
--- a/applicationinsights.json
+++ b/applicationinsights.json
@@ -16,9 +16,6 @@
   "selfDiagnostics": {
     "destination": "console"
   },
-  "sampling": {
-    "percentage": 100
-  },
   "preview": {
     "sampling": {
       "overrides": [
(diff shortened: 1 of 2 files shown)


applicationinsights.json
```diff
@@ -6,13 +6,19 @@
...
+      "enabled": false
+    },
     "logging": {
       "level": "DEBUG"
     }
   },
   "selfDiagnostics": {
     "destination": "console"
   },
+  "sampling": {
```
petergphillips, line 19: this will grab all requests, not just a sample of them.

applicationinsights.json
```diff
@@ -6,13 +6,19 @@
     "service.version": "${BUILD_NUMBER}"
   },
   "instrumentation": {
+    "springScheduling": {
```
petergphillips, line 9: don't think this will make any difference as you're up to 3.4.10 app insights already

src/main/resources/logback-spring.xml
```diff
@@ -25,34 +25,31 @@
 
     <logger name="uk.gov.justice.digital.hmpps.visitscheduler.VisitSchedulerApplicationKt" additivity="false"
             level="DEBUG">
-        <appender-ref ref="logAppender"/>
         <appender-ref ref="consoleAppender"/>
     </logger>
 
     <logger name="uk.gov.justice.digital.hmpps" additivity="false" level="DEBUG">
-        <appender-ref ref="logAppender"/>
+        <appender-ref ref="consoleAppender"/>
```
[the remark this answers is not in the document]
AledWEvans: This has never been correct, I would just have a consoleAppender and not use springProfile at all in log back. But not sure this is the cause of your problem as this has only just started happening.

applicationinsights.json
AledWEvans: Could this be the problem?

applicationinsights.json
dhirajshettymoj: will remove this bit.

applicationinsights.json
dhirajshettymoj: based it on [the](https://github.com/ministryofjustice/hmpps-restricted-patients-api/blob/main/applicationinsights.json). Should I remove the sampling bit altogether?

src/main/resources/logback-spring.xml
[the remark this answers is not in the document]
dhirajshettymoj: do you mean the logAppender?

AledWEvans (commented):
LGTM

AledWEvans (approved):
LGTM

PR #328, opened 2023-04-06
vb 2083 allocate capacities by housing level in a session

AledWEvans:

Allows different capacities to be configured for location groups for the same session times. 

To allow us to configure for prisons such as BNI


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
```diff
@@ -102,6 +104,26 @@ class MigrateVisitService(
...
+        val sessionCapacity = sessionService.getSessionCapacity(location, migrateVisitRequest.prisonCode, sessionDate, startTime, endTime)
+        return sessionCapacity?.capacityGroup
+      } else {
+        val sessionCapacities = sessionService.getSessionCapacity(migrateVisitRequest.prisonCode, sessionDate, startTime, endTime)
+        // TODO this is a list > how do we do this?
+      }
+    }
+    return null
+  }
+
```
AledWEvans, line 126: The above code is a tricky one, I think we should have a manual solution to this some how because migration is a pain

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
```diff
@@ -180,7 +180,7 @@ class VisitSessionController(
       example = "14:30:00",
     )
     sessionEndTime: LocalTime,
-  ): SessionCapacityDto {
+  ): List<SessionCapacityDto> {
```
AledWEvans, line 183: @tpmcgowan and @hutcheonb-moj this will be interesting because you can now get more that one SessionCapacityDto back... not sure how this is going to work in the front end!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/config/VisitSchedulerExceptionHandler.kt
```diff
@@ -74,6 +75,18 @@ class VisitSchedulerExceptionHandler(
     return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error)
   }
 
+  @ExceptionHandler(MigratedVisitCapacityGroupMatchException::class)
+  fun handleMigratedVisitCapacityGroupMatchException(e: Exception): ResponseEntity<ErrorResponse> {
+    log.debug("Migration exception: {}", e.message)
+    val error = ErrorResponse(
+      status = HttpStatus.BAD_REQUEST,
+      userMessage = "Migration failure: could not find matching session template",
```
psoleckimoj, line 83: Is this message correct? The handler talks about visit capacity group?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
```diff
@@ -102,6 +116,151 @@ class MigrateVisitService(
...
+      val sortedByLevelMatch = templateLocationMap.toList().sortedWith(proximityComparator)
+      return sessionLocationTemplates.find { it.reference == sortedByLevelMatch.first().first }!!
+    }
+    return sessionTemplates.first()
+  }
+
+  private fun createTemplateLocationMap(sessionLocationTemplates: List<SessionTemplate>): MutableMap<String, MutableList<PermittedSessionLocation>> {
+    val templateLocationMap = mutableMapOf<String, MutableList<PermittedSessionLocation>>()
+    sessionLocationTemplates.forEach { st ->
```
psoleckimoj, line 255: `st` and `lg` should be expanded, picky I know 😄

src/main/resources/db/migration/V2_20__rename_visit_room.sql
```diff
@@ -0,0 +1,19 @@
...
+
+UPDATE visit
+    SET visit_room = NULL
+        FROM (SELECT v.id,v.visit_room FROM visit v JOIN legacy_data ld ON ld.visit_id = v.id) AS subquery
+    WHERE visit.id = subquery.id;
+
+
+ALTER TABLE visit RENAME COLUMN visit_room TO capacity_group;
+ALTER TABLE session_template RENAME COLUMN visit_room TO capacity_group;
+ALTER TABLE session_template ALTER COLUMN capacity_group drop not null;
```
psoleckimoj, line 19: Missing end newline

psoleckimoj (approved):
LGTM, just a couple of comments 🙂

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/config/VisitSchedulerExceptionHandler.kt
AledWEvans: just generic message to customer, dev message will have the detail

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
AledWEvans: I don't think it's picky at all, I hate using st or lg,.. it's not self descriptive, will change..

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/config/VisitSchedulerExceptionHandler.kt
AledWEvans: generic message to user, dev message has detail

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/ReserveVisitSlotDto.kt
```diff
@@ -20,9 +20,8 @@ data class ReserveVisitSlotDto(
   @Schema(description = "Prison Id", example = "MDI", required = true)
   @field:NotBlank
   val prisonCode: String,
-  @Schema(description = "Visit Room", example = "A1", required = true)
-  @field:NotBlank
-  val visitRoom: String,
+  @Schema(description = "Capacity group", example = "Main Group", required = false)
```
tpmcgowan, line 23: So, looks like we're repurposing the `visitRoom` as `capacityGroup`. Wondering if we actually need to pass this to/from front-end at all (e.g. like here when reserving a slot)? We don't do anything with `visitRoom` in the front-end. Are there requirements as part of this for us to be doing something with `capacityGroup`?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
AledWEvans: interesting question??

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/ReserveVisitSlotDto.kt
tpmcgowan: As discussed - we do in fact need it!

tpmcgowan (approved):
LGTM - have raised issue of displaying capacity with Simon and filed a couple of tickets for updating orchestration & front-end

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
tpmcgowan: Simon's responded to say just display a single total capacity - so I guess either merge them into one `SessionCapacityDto` here or we add up multiple ones on the front-end. Probably simpler for you to add and just return one 🤷‍♂️?

PR #334, opened 2023-04-18
Vb 2139 non associations are not being considered in booking process

dhirajshettymoj:

Logs any non associations for the next n days

To see if any non association visits have been allowed.


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/task/VisitTask.kt
```diff
@@ -36,4 +46,31 @@ class VisitTask(
...
+          startDateTime = LocalDateTime.now().plusDays(i).with(LocalTime.MIN),
+          endDateTime = LocalDateTime.now().plusDays(i).with(LocalTime.MAX),
+        )
+        val visits = visitService.findVisitsByFilterPageableDescending(visitFilter)
+
+        visits.forEach {
+          val sessions = sessionService.getVisitSessions(it.prisonCode, it.prisonerId, i, i)
+          if (sessions.isEmpty()) {
+            log.info("Visit with reference - ${it.reference} ,prisoner id - ${it.prisonerId}, prison id - ${it.prisonerId}, start time - ${it.startTimestamp}, end time - ${it.endTimestamp} flagged for check - possible non associations on same day.")
+          }
```
AledWEvans, line 71: May be delay between each iteration,... Thread.sleep()?

dhirajshettymoj: Added sleep of 1/2 second between visits and updated log message.

AledWEvans: would also do a try catch 

`    try {
      Thread.sleep(500)
    } catch (e : InterruptedException) {
      log.error("Sleep failed : $e")
    }`

AledWEvans (approved):
LGTM

PR #339, opened 2023-04-18
Connect using clientid 2

dhirajshettymoj:

Connects using client id

2nd attempt after adding properties to values.yaml


AledWEvans (approved):
LGTM - makes sense now, with helm values, sorry did not think they were missing

PR #341, opened 2023-04-21
Remove list of capacity objects from api

AledWEvans:

Removes multiple capacity dto's being returned when different capacity groups, it now adds them all up

Reduce UI change


tpmcgowan (approved):
LGTM!

PR #346, opened 2023-04-24
Still had capacoty_group in temp table

AledWEvans:

Small error in sql script capacoty_group was still in sql script re verted back to visit_room

Sql would not run in preprod other wise


tpmcgowan (approved):
LGTM

PR #349, opened 2023-04-26
changed capacity group name back to original visit room name, also ad…

AledWEvans:

reverts rooms names from Capacity_Main_Group to "Visits Main Hall"

SYCON needs room names for NART reporting.


tpmcgowan (approved):
LGTM

PR #353, opened 2023-05-03
A_HIGH should not be on FNI_NON_CAT_A_HIGH category grouping.

dhirajshettymoj:

Corrects the SQL script as A_HIGH was wrongly configured in Non CAT A High grouping.

Data correction on Dev env.


AledWEvans (approved):
LGTM

PR #354, opened 2023-05-05
Added new logic for mapping visits to session templates

AledWEvans:

It make sure we migrate dat correctly so we can have accurate capacity counts

To make sure migrated data is accurate in the 28 period


dhirajshettymoj (approved):
Looks good to merge. We will need to do a Preprod migration run to test live scenarios.

PR #356, opened 2023-05-09
Update MigrateVisitService.kt

AledWEvans:
Change future date to use just Date not Time and Date

Connects session template on the date rather than date and time

to improve availability counts


dhirajshettymoj (approved):
LGTM.

PR #371, opened 2023-05-16
VB-2242 When Migration a visit for prisoner A1066EW for Bristol BLI W…

AledWEvans:
…ING D wrong session is selected - FIX, with a little tidy up of code.

It fixes the problem with location mapping to the session template, previously it compare all matching location in a match group now it match matching location only

to fix the about bug, so our session mapping is improved between migrate visits and session template


dhirajshettymoj (approved):
LGTM

PR #374, opened 2023-05-16
Added new logs to help test when a prisoner is moved

AledWEvans:
added extra logs because some time we have prisoners who have moved and need to know about this during migration.


dhirajshettymoj (approved):
LGTM

PR #376, opened 2023-05-17
Added new logs to help test when a prisoner is moved, error in logs

AledWEvans:
corrected a small error in logs


dhirajshettymoj (approved):
LGTM

PR #377, opened 2023-05-17
Added new logs to help test when a prisoner is moved, error in logs

AledWEvans:
extra bit of info on exception message for migrated daata


dhirajshettymoj (approved):
LGTM

PR #378, opened 2023-05-17
Updated BNI timetable to kick in after 28 days. Time in csv file was …

AledWEvans:
Time table needed to be changed for production, for BNI, the purpose of this is to prevent table counts being out of wack!


dhirajshettymoj (approved):
LGTM

PR #383, opened 2023-05-18
VB-2246 - updates to Durham timetables.

dhirajshettymoj:

Updates Durham timetable.

Bring Durham timetable in line with existing timetable.


dhirajshettymoj:
closed the PR for now. But if we think we need to update sql generator will get reopen

PR #385, opened 2023-05-19
Added migration date and time to help in recovery of failed migrations.

AledWEvans:
Added migration date and time to help in recovery of failed migration to default to null


dhirajshettymoj (approved):
LGTM

PR #389, opened 2023-05-22
Updated failed test based on a fix date for bi weekly

AledWEvans:
Test was failing for bi weekly because the date used was based on a fixed date, so it was passing one week failing the next


dhirajshettymoj (approved):
LGTM

PR #390, opened 2023-05-22
VB-2246 - updates to Durham session templates to match current timetable

dhirajshettymoj:
VB-2246 - updates to Durham session templates to match current timetable

Updates to Durham timetable to bring it in Live with Lve.

Timetable updates for Durham prison


AledWEvans (approved):
LGTM

PR #409, opened 2023-06-01
VB-1295 - interim fix to reduce time span to 10 minutes

dhirajshettymoj:
VB-1295 - interim fix to reduce the time span from 20 minutes to 10 minutes. Any reserved visits older than 10 minutes will be deleted.

Deletes reserved visits older than 10 minutes instead of the current 20 minutes.

Interim fix as call centre cannot book visits for same prisoner again for 20 minutes. So reducing the time to 10 minutes.


hutcheonb-moj (approved):
LGTM

hutcheonb-moj (approved):
LGTM

PR #414, opened 2023-06-05
Vb 2297 administration service add remove exclude dates to / from prison

dhirajshettymoj:

Admin endpoint to add / remove exclude dates to / from prison

Updates to the administration service.


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/PrisonExcludeDateRepository.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.PrisonExcludeDate
+import java.time.LocalDate
+
+@Repository
+interface PrisonExcludeDateRepository : JpaRepository<PrisonExcludeDate, Long> {
+  @Modifying
+  @Query(
+    "delete  FROM PrisonExcludeDate ped " +
+      "WHERE ped.prisonId = :prisonId" +
```
AledWEvans, line 15: could do a join rather with prison code rather than use prisonId

AledWEvans: I see why you have done this now, no worries leave it as id...

AledWEvans (approved):
LGTM

PR #417, opened 2023-06-06
A fix to ensure schedule in correct time order

AledWEvans:

A fix to ensure schedule in correct time order

to improve customer exsperiance


dhirajshettymoj (approved):
LGTM

PR #431, opened 2023-06-09
Fix to resolve issues with multiple events being created

AledWEvans:

This pull requests moves the update of the visit entity for booking and cancellation into a helper class, so the transaction can be complete before the domain events are sent (Domain events can take some time). Therefore other threads with the same action and details will use the latest persisted data. Hopefully duplicate requests will be caught by isApplicationBooked or isBookingCancelled and the existing details will be return with no additional process.   

To prevent duplicate events from being sent to the domain events and down stream actions will not be actioned.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -10,6 +10,7 @@ import org.springframework.data.domain.PageRequest
 import org.springframework.data.domain.Pageable
 import org.springframework.data.domain.Sort
 import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Propagation.NEVER
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.visitscheduler.dto.CancelVisitDto
 import uk.gov.justice.digital.hmpps.visitscheduler.dto.ChangeVisitSlotRequestDto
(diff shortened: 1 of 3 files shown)


AledWEvans:
I'm still wondering if this is the correct place and approach to solve this issue... SNS and SQS should be send and forget and should not delay the process. Should Push or Put should have retries?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -229,6 +230,7 @@ class VisitService(
     )
   }
 
+  @Transactional(propagation = NEVER)
```
andymarke, line 233: Are you sure this means "don't start a transaction for this method"? Spring documentation suggests it means something different - but I'm not sure since I have never used it before?

(obviously you could remove the class annotation and add one to each method except this one)

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
AledWEvans: My understand of NEVER is that it checks to see if there is an existing transaction and if so throws an Illegal state exception. So by adding this to the method I'm preventing the use of this method within an existing transaction,.. Integration test would fail. The call to persistVisitHelper then creates a new transaction.

PR #434, opened 2023-06-13
VB-2335 Create active in active get Session Template end point

AledWEvans:
Changes session template end point to allow UI to select session templates by from and to date using a flag

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionAdminController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionAdminController.kt
@@ -23,16 +23,21 @@ import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.visitscheduler.config.ErrorResponse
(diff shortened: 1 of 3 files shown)


dhirajshettymoj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionAdminController.kt
```diff
@@ -71,28 +76,23 @@ class VisitSessionAdminController(
...
-    rangeEndDate: LocalDate?,
+    rangeType: SessionTemplateRangeType,
   ): List<SessionTemplateDto> {
-    return sessionTemplateService.getSessionTemplates(prisonCode, dayOfWeek, rangeStartDate, rangeEndDate)
+    var fromDate: LocalDate? = null
+    var toDate: LocalDate? = null
+    if (ACTIVE_OR_FUTURE == rangeType) {
+      fromDate = LocalDate.now()
+    }
+    if (HISTORIC == rangeType) {
```
tpmcgowan, line 92: If choosing `HISTORIC`, what should I expect? Should it be templates that are no longer active at all?

Having tested, it is returning templates where the **from/to** dates are **both** in the past. But it is also returning templates that don't have a `dateTo` set at all - i.e. their `fromDate` is in the past but they are still active.

AledWEvans: @tpmcgowan Yes you would get one's that are no longer valid

PR #441, opened 2023-06-14
Vb 2341 move admin endpoints into correct code base/package

AledWEvans:
Re factored end point's no functional change, also rename classes and moved to improved packages


dhirajshettymoj (approved):
LGTM

PR #446, opened 2023-06-19
Vb 2321 delete api function on session templates add validation

AledWEvans:

add's validation to delete end point, if session template has visits then it cannot be deleted.

to make sure session templates can only be delete when there is no associated visits, to prevent live session templates being deleted

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
@@ -84,6 +84,14 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
   ): Boolean
 
   @Query(
+    "SELECT count(v) > 0 FROM Visit v " +
+      "WHERE v.sessionTemplateReference = :sessionTemplateReference",
+  )
+  fun hasVisitsForSessionTemplate(
+    sessionTemplateReference: String,
+  ): Boolean
+
+  @Query(
     "SELECT v.visitRestriction AS visitRestriction, COUNT(v) AS count  FROM Visit v " +
       "WHERE v.prison.code = :prisonCode AND " +
       "v.visitStart >= :startDateTime AND " +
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -83,6 +83,14 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
     endDateTime: LocalDateTime,
   ): Boolean
 
+  @Query(
+    "SELECT count(v) > 0 FROM Visit v " +
+      "WHERE v.sessionTemplateReference = :sessionTemplateReference",
```
dhirajshettymoj, line 88: need to add status as BOOKED

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/admin/AdminDeleteSessionTemplateTest.kt
```diff
@@ -63,4 +64,24 @@ class AdminDeleteSessionTemplateTest(
     Assertions.assertThat(testSessionLocationGroupRepository.hasJoinTable(sessionTemplate.id, sessionGroup1.id)).isFalse
     Assertions.assertThat(testSessionLocationGroupRepository.hasJoinTable(sessionTemplate.id, sessionGroup2.id)).isFalse
   }
+
+  @Test
+  fun `cannot delete session template by reference with existing visits test`() {
```
dhirajshettymoj, line 69: could add another test for non booked visits post adding status on query

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
dhirajshettymoj: resolved after having a chat with @AledWEvans as this cannot be done once the session template goes Live and is used to create visits whatever the Status.

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/admin/AdminDeleteSessionTemplateTest.kt
dhirajshettymoj: resolved after having a chat with @AledWEvans as this cannot be done once the session template goes Live and is used to create visits whatever the Status.

dhirajshettymoj (approved):
LGTM post discussions with Aled around visitStatus.

PR #447, opened 2023-06-19
VB-2346 Update visit schedule for HMP Wandsworth (WWI) in VSIP

AledWEvans:
Updates from Wandsworth to ensure time table excludes first Wednesday of every month plus new time slot for Wednesday


src/main/resources/db/migration/data/R__add_prison_codes.sql
```diff
@@ -32,12 +32,41 @@ INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-12-22'
...
 INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-06-15' from prison where code = 'PNI';
 INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-07-18' from prison where code = 'PNI';
 INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-09-19' from prison where code = 'PNI';
 INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-10-17' from prison where code = 'PNI';
 INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-11-21' from prison where code = 'PNI';
-INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-12-12' from prison where code = 'PNI';
\ No newline at end of file
+INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-12-12' from prison where code = 'PNI';
+
+-- HMP Durham (PNI)
```
dhirajshettymoj, line 45: minor on comment - prison code to be updated

src/main/resources/db/migration/data/R__add_prison_codes.sql
```diff
@@ -32,12 +32,41 @@ INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-12-22'
...
+INSERT INTO prison_exclude_date(prison_id, exclude_date) Select id,'2023-06-22' from prison where code = 'DMI';
```
dhirajshettymoj, line 46: do we need to exclude this date for DMI? Is this part of a different ticket?

AledWEvans: It's because this SQL statement VB_2274_exclude_durham_22_Jun.sql was added to the wrong place so I move it in here and deleted the old one,... I think this was done when I was on holiday,... if it's not in this file it will be deleted..

PR #451, opened 2023-06-22
VB-2358 Introduce logic to server side - n weekly

AledWEvans:

- Introduce logic to server side
- Return results for display in the UI
- Create SQL scripts to migrate existing sessions that are not weekly
- Migrate logic
- Updates SessionScheduleDto to  use SessionTimeSlot and DateRangeDto

To allow n weekly to work, so that complex time tables can be catered for,... also refactor SessionScheduleDto to standardise approach.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/sessions/SessionTemplateDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/sessions/SessionTemplateDto.kt
@@ -27,8 +27,8 @@ data class SessionTemplateDto(
   val sessionDateRange: SessionDateRangeDto,
   @Schema(description = "visit type", example = "SOCIAL", required = true)
   val visitType: VisitType,
-  @Schema(description = "biWeekly", example = "true", required = true)
-  val biWeekly: Boolean,
+  @Schema(description = "number of weeks until the weekly day is repeated", example = "1", required = true)
+  val weeklyFrequency: Int,
   @Schema(description = "Visit Room", example = "A1 L3", required = true)
   @field:NotBlank
   val visitRoom: String,
@@ -56,6 +56,6 @@ data class SessionTemplateDto(
     permittedLocationGroups = sessionTemplateEntity.permittedSessionLocationGroups.map { SessionLocationGroupDto(it) },
     prisonerCategoryGroups = sessionTemplateEntity.permittedSessionCategoryGroups.map { SessionCategoryGroupDto(it) },
     prisonerIncentiveLevelGroups = sessionTemplateEntity.permittedSessionIncentiveLevelGroups.map { SessionIncentiveLevelGroupDto(it) },
-    biWeekly = sessionTemplateEntity.biWeekly,
+    weeklyFrequency = sessionTemplateEntity.weeklyFrequency,
   )
 }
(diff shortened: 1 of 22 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/sessions/SessionTemplateDto.kt
```diff
@@ -27,8 +27,8 @@ data class SessionTemplateDto(
   val sessionDateRange: SessionDateRangeDto,
   @Schema(description = "visit type", example = "SOCIAL", required = true)
   val visitType: VisitType,
-  @Schema(description = "biWeekly", example = "true", required = true)
-  val biWeekly: Boolean,
+  @Schema(description = "number of weeks until the weekly day is repeated", example = "1", required = true)
+  val weeklyFrequency: Int,
```
dhirajshettymoj, line 31: we could add a @Min(0) validation to avoid lower than 0 frequency or change variable to UInt. The validation might be better option. Also is weeklyFrequency of 0 allowed?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/SessionTemplateRepository.kt
```diff
@@ -119,6 +119,6 @@ interface SessionTemplateRepository : JpaRepository<SessionTemplate, Long> {
   fun updateOpenCapacityByReference(reference: String, openCapacity: Int): Int
 
   @Modifying
-  @Query("Update SessionTemplate s set s.biWeekly = :biWeekly WHERE s.reference = :reference")
-  fun updateBiWeeklyByReference(reference: String, biWeekly: Boolean): Int
+  @Query("Update SessionTemplate s set s.weeklyFrequency = :weeklyFrequency WHERE s.reference = :reference")
+  fun updateBiWeeklyByReference(reference: String, weeklyFrequency: Int): Int
```
dhirajshettymoj, line 123: the method name can be updated from updateBiWeekly to updateWeeklyFrequency

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/utils/SessionDatesUtil.kt
```diff
@@ -17,54 +17,50 @@ class SessionDatesUtil {
...
 
-  fun isBiWeeklySkipDate(
-    validFromDate: LocalDate,
+  fun isWeeklySkipDate(
     sessionDate: LocalDate,
-  ) = WEEKS.between(validFromDate, sessionDate).toInt() % 2 != 0
+    sessionTemplate: SessionTemplate,
+  ): Boolean {
+    val validFromMonday = getValidFromMonday(sessionTemplate)
+    return WEEKS.between(validFromMonday, sessionDate).toInt() % sessionTemplate.weeklyFrequency != 0
```
dhirajshettymoj, line 64: can weeklyFrequency be 0 if it doesn't repeat? If it can be 0 this will throw ArithmeticException for divide by 0

tpmcgowan:
Have looked specifically at the DTO changes that will affect staff UI. These look fine; I'll leave Dhiraj to approve 🙂

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/sessions/SessionTemplateDto.kt
AledWEvans: It would be @min(1)

dhirajshettymoj (approved):
LGTM

PR #452, opened 2023-06-22
VB-2377 Remove dates from Foston Hall schedule in VSiP (FHI)

AledWEvans:
Remove dates from Foston Hall schedule in VSiP (FHI)


hutcheonb-moj (approved):
LGTM

PR #456, opened 2023-06-28
VB-2386 VB-2386 Block date from Durham & Cardiff schedule

AledWEvans:
VB-2388 Block date from Cardiff schedule in VSiP (CFI)
VB-2386 Block date from Durham schedule in VSiP (DMI)


dhirajshettymoj (approved):
LGTM

PR #460, opened 2023-06-29
VB-2395 Block dates from Cardiff schedule in VSiP (CFI)

AledWEvans:
Description

We need to remove the following dates from the social visits timetable for Cardiff:

12 7 2023 

19 7 2023


hutcheonb-moj (approved):
LGTM

PR #462, opened 2023-06-30
Refactor of exclude date, after review from javier

AledWEvans:
Refactor of exclude date, after review from @infojav


infojav (approved):

dhirajshettymoj (approved):
LGTM

PR #467, opened 2023-07-05
Vb 2056 update api dummy out resp requ

AledWEvans:
added dummy API so UI guys can get on with work

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/VisitDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/VisitDto.kt
@@ -64,7 +64,7 @@ data class VisitDto(
   val modifiedTimestamp: LocalDateTime,
   @Schema(description = "What was the last application method", required = true)
   @field:NotNull
-  val lastApplicationMethod : ApplicationMethod
+  val lastApplicationMethod: ApplicationMethod,
 ) {
   constructor(visitEntity: Visit) : this(
     lastApplicationMethod = ApplicationMethod.UNKNOWN,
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/VisitDto.kt
```diff
@@ -60,9 +62,12 @@ data class VisitDto(
   @Schema(description = "The visit modified date and time", example = "2018-12-01T13:45:00", required = true)
   @field:NotBlank
   val modifiedTimestamp: LocalDateTime,
-
+  @Schema(description = "What was the last application method", required = true)
+  @field:NotNull
+  val lastApplicationMethod: ApplicationMethod,
```
AledWEvans, line 67: Do we need this @hutcheonb-moj? would rather not include it, if front dose not use it?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -271,7 +272,7 @@ class VisitService(
   @Deprecated("This method has been deprecated.")
```
dhirajshettymoj, line 272: is this method needed anymore?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -271,7 +272,7 @@ class VisitService(
...
   fun cancelVisit(reference: String, outcomeDto: OutcomeDto): VisitDto {
     LOG.debug("Enter legacy call to cancelVisit $reference")
-    val cancelVisitDto = CancelVisitDto(outcomeDto, NOT_KNOWN_NOMIS)
+    val cancelVisitDto = CancelVisitDto(outcomeDto, NOT_KNOWN_NOMIS, UNKNOWN)
```
dhirajshettymoj, line 275: instead of UNKONWN can ApplicationMethod be NOMIS here?

AledWEvans: This is just to dummy out next ticket will put in correct value..

dhirajshettymoj (approved):
LGTM

PR #468, opened 2023-07-05
Fix up create group dto endpoint

AledWEvans:
RequestBody for update and create location groups was incorrect and update on category


tpmcgowan (approved):
LGTM!

PR #478, opened 2023-07-13
Admin controllers delete return

hutcheonb-moj:

Similar to https://github.com/ministryofjustice/visit-scheduler/pull/463
Return JSON when delete is successful

Copy the work done in the above mentioned PR for:
Incentive groups
Location groups
Category groups

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/admin/LocationGroupAdminController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/admin/LocationGroupAdminController.kt
@@ -178,7 +178,7 @@ class LocationGroupAdminController(
     reference: String,
   ): ResponseEntity<String> {
     sessionTemplateService.deleteSessionLocationGroup(reference)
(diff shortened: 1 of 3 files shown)


AledWEvans (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/admin/LocationGroupAdminController.kt
```diff
@@ -144,7 +146,7 @@ class LocationGroupAdminController(
   }
 
   @PreAuthorize("hasRole('VISIT_SCHEDULER_CONFIG')")
-  @DeleteMapping(REFERENCE_LOCATION_GROUP_ADMIN_PATH)
+  @DeleteMapping(REFERENCE_LOCATION_GROUP_ADMIN_PATH, produces = [MediaType.APPLICATION_JSON_VALUE])
```
AledWEvans, line 149: don't normally use this, but I guess I normally return a DTO so it knows it's JSON

PR #479, opened 2023-07-14
Vb 2458 admin api create groups failing

dhirajshettymoj:

Fix for issue where in group id was being allocated a 0 when adding a permitted location.

Admin API fixes


AledWEvans (approved):
LGTM

PR #484, opened 2023-07-17
VB-2470 add validation around updates of valid to date

dhirajshettymoj:
Validations and test around updates of valid to date when updating a session template.

Admin API improvements.


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -91,6 +91,16 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
...
+  @Query(
+    "SELECT count(v) > 0 FROM Visit v " +
+      "WHERE v.sessionTemplateReference = :sessionTemplateReference AND " +
+      "v.visitStart >= :visitsFromDate",
+  )
+  fun hasVisitsForSessionTemplateFromDate(
+    sessionTemplateReference: String,
+    visitsFromDate: LocalDateTime,
+  ): Boolean
+
```
AledWEvans, line 103: `SessionTemplateRepository.getVisitCount`
I wonder if we should scrap one of these methods,...

AledWEvans: In both cases we use `v.visit_start` should we be using the template start and end date?

dhirajshettymoj: Thanks @AledWEvans . I have removed hasVisitsForSessionTemplateFromDate and refactored hasVisitsForSessionTemplate to use validFromDate.

AledWEvans (approved):
LGTM

PR #486, opened 2023-07-18
Vb 2315 add custom event to app insights for when a reservation is released

AledWEvans:
Add customEvent to App Insights for when a reservation is released


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -48,6 +48,7 @@ import uk.gov.justice.digital.hmpps.visitscheduler.repository.SupportTypeReposit
 import uk.gov.justice.digital.hmpps.visitscheduler.repository.VisitRepository
 import uk.gov.justice.digital.hmpps.visitscheduler.service.TelemetryVisitEvents.VISIT_BOOKED_EVENT
 import uk.gov.justice.digital.hmpps.visitscheduler.service.TelemetryVisitEvents.VISIT_CANCELLED_EVENT
+import uk.gov.justice.digital.hmpps.visitscheduler.service.TelemetryVisitEvents.VISIT_DELETED_EVENT
```
dhirajshettymoj, line 51: should the VISIT_DELETED_EVENT be updated to visit-slot-released as per the ticket?

dhirajshettymoj (commented):
minor comment added, looks good otherwise

dhirajshettymoj (approved):
LGTM

PR #493, opened 2023-07-25
Remove exclude dates and add prison script

AledWEvans:
Remove code that sets prisons and exclude dates... to DB on startup because we are now using the admin tool


dhirajshettymoj (approved):
LGTM

PR #494, opened 2023-07-26
VB-2071 Add NOT_APPLICABLE application method type

tpmcgowan:

Adds new value `NOT_APPLICABLE` to application method type.

Capturing application method type is for requests by visitors. This new NOT_APPLICABLE value is for cases where, for example, a visit might be cancelled by the establishment itself.


dhirajshettymoj (approved):
LGTM

PR #502, opened 2023-08-02
Vb 2592 calculate the cost of our service migration event date is not migrating properly

AledWEvans:

Fixes a bug described in Vb 2592


dhirajshettymoj (approved):
LGTM

PR #503, opened 2023-08-02
VB-2592 Calculate the cost of our service : Migration Event Date is n…

AledWEvans:
To fix bug detected in the migration process.


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
```diff
@@ -126,12 +127,17 @@ class MigrateVisitService(
...
-      eventAuditRepository.updateCreateTimestamp(it, eventAudit.id)
+      if (migrateVisitRequest.visitStatus == BOOKED) {
+        eventAuditRepository.updateCreateTimestamp(it, eventAudit.id)
+      }
     }
 
     // Do this at end of this method, otherwise modify date would be overridden
     migrateVisitRequest.modifyDateTime?.let {
       visitRepository.updateModifyTimestamp(it, visitEntity.id)
+      if (migrateVisitRequest.visitStatus == CANCELLED) {
```
AledWEvans, line 138: had to use status in the end as migration always gave me both values

dhirajshettymoj (approved):
LGTM

PR #507, opened 2023-08-07
delete visits that are more than 6 months in the future.

dhirajshettymoj:

Script to delete future dated visits.

Delete older visits.

### DIFF
--- a/src/main/resources/db.scripts.mvp/delete_migrated_visits_more_than_6_months_in_future.sql
+++ b/src/main/resources/db.scripts.mvp/delete_migrated_visits_more_than_6_months_in_future.sql
@@ -1,0 +1,27 @@
+BEGIN;
+
+SET SCHEMA 'public';
+
+    CREATE TEMP TABLE tmp_future_visits_to_be_deleted(
+    visit_id         int
+);
+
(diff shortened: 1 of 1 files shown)


rj-adams:
Surely, the intent is to delete **_too far in the future_** visits - not **_older_** visits.

src/main/resources/db.scripts.mvp/delete_migrated_visits_more_than_6_months_in_future.sql
```diff
@@ -0,0 +1,27 @@
+BEGIN;
+
```
AledWEvans, line 2: This work has already been done, 

R_remove_legacy_visits_that_are_six_months_in_future.sql

Why not just use that.

rj-adams: Or better - prevent them from being migrated in the first place - return a error if visit dates are too far in the future.

AledWEvans: Other Script has been run

PR #508, opened 2023-08-07
Updated visit booked/reserved count check

AledWEvans:
To fix a booking count issues, it now uses the session template start and end time to work out the booking count


dhirajshettymoj (approved):
LGTM

PR #509, opened 2023-08-08
VB-2651 Missing room name on update session template

AledWEvans:
VB-2651 Missing room name on update session template

Fixes tech defect, missing room update for session template


hutcheonb-moj (approved):
LGTM

PR #512, opened 2023-08-09
VB-2658 Performance issue in live, after release

AledWEvans:
Performance issues in live, after https://dsdmoj.atlassian.net/browse/VB-2647
adding index to improve speed.


dhirajshettymoj (approved):
LGTM. As discussed the file rename should not have an issue.

PR #513, opened 2023-08-09
Extra performance improvements for book count.

AledWEvans:
Extra performance improvements as casing prevents indexing.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
@@ -14,7 +14,6 @@ import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.Visit
 import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.projections.VisitRestrictionStats
 import java.time.LocalDate
 import java.time.LocalDateTime
-import java.time.LocalTime
 
 @Repository
 interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor<Visit> {
@@ -105,8 +104,6 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
       "WHERE p.code = :prisonCode AND " +
       "v.session_template_reference = :sessionTemplateReference AND " +
       "(v.visit_start >= :sessionDate AND v.visit_start < (CAST(:sessionDate AS DATE) + CAST('1 day' AS INTERVAL))) AND " +
-      "st.start_time = :sessionStartTime AND " +
-      "st.end_time = :sessionEndTime AND " +
       "v.visit_restriction in ('OPEN','CLOSED') AND " +
       "v.visit_status = 'BOOKED' " +
       "GROUP BY v.visit_restriction",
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -104,7 +104,7 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
       "JOIN session_template st ON st.reference = v.session_template_reference " +
       "WHERE p.code = :prisonCode AND " +
       "v.session_template_reference = :sessionTemplateReference AND " +
-      "Date(v.visit_start) = :sessionDate AND " +
+      "(v.visit_start >= :sessionDate AND v.visit_start < CAST(:sessionDate AS DATE) + CAST('1 day' AS INTERVAL)) AND " +
```
dhirajshettymoj, line 107: understand more after the call we had. Just a minor comment could we put can this statement in brackets - CAST(:sessionDate AS DATE) + CAST(‘1 day’ AS INTERVAL)

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -125,7 +125,7 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
       "JOIN prison p ON p.id = v.prison_id " +
       "JOIN session_template st ON st.reference = v.session_template_reference " +
       "WHERE p.code = :prisonCode AND " +
-      "Date(v.visit_start) = :sessionDate AND " +
+      "(v.visit_start >= :sessionDate AND v.visit_start < CAST(:sessionDate AS DATE) + CAST('1 day' AS INTERVAL)) AND " +
```
dhirajshettymoj, line 128: Also the sessionStartTime and endtime check might not be needed anymore.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
```diff
@@ -125,7 +125,7 @@ interface VisitRepository : JpaRepository<Visit, Long>, JpaSpecificationExecutor
...
       "st.start_time = :sessionStartTime AND " +
       "st.end_time = :sessionEndTime AND " +
```
dhirajshettymoj, line 130: Also the sessionStartTime and endtime check might not be needed anymore.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
rj-adams: Would be good to see what the query planner does with this query - might point us towards ways we can improve it further

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitRepository.kt
AledWEvans:
```
"st.start_time = :sessionStartTime AND " +
"st.end_time = :sessionEndTime AND " +
```
Is key to the logic I dont understand why we would remove it?

AledWEvans: Your right, all I need is the session template ref

dhirajshettymoj (approved):
LGTM

PR #522, opened 2023-09-01
VB-2715 added end point with dummy response

AledWEvans:
Implement non-association for SQS - create listener in orchestration, added endpoint to allow orchestration work to be complete

It adds an end point with dummy response.

to allow use to continue separate threads of coding work..


dhirajshettymoj (approved):
LGTM

PR #527, opened 2023-09-07
VB-2711 Implement non-association for SQS - storage

AledWEvans:
Vb 2711 implement non association for sqs   storage


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationService.kt
```diff
@@ -39,10 +41,23 @@ class VisitNotificationService(
     }
   }
 
+  private fun isNotADeleteEvent(nonAssociationChangedNotification: NonAssociationChangedNotificationDto): Boolean {
+    val nonAssociations = prisonerService.getOffenderNonAssociationList(nonAssociationChangedNotification.prisonerNumber)
+    return nonAssociations.any { it.offenderNonAssociation.offenderNo == nonAssociationChangedNotification.nonAssociationPrisonerNumber }
+  }
```
AledWEvans, line 47: Should I also check the dates, just dont know if they will exisit on a delete event,.. only one way to find out here,... also I dont know if there will be a race condition between this event updating the prison service and doing the update,... here... only one way to find out

rj-adams: Visit notification service should be listening for the booking create/update/delete events which have been generated by the visit scheduler - then sending out notifications accordingly.

Visit notifications should be an independent service - not part of the visit scheduler!

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
```diff
@@ -1,8 +1,23 @@
...
+import uk.gov.justice.digital.hmpps.visitscheduler.service.NotificationEventType
 
 @Repository
-interface VisitNotificationEventRepository : JpaRepository<VisitNotificationEvent, Int>
+interface VisitNotificationEventRepository : JpaRepository<VisitNotificationEvent, Int> {
+
+  @Query(
+    "SELECT count(ve) > 0" +
+      " FROM visit_notification_event ve " +
+      " WHERE ve.create_timestamp BETWEEN NOW() - INTERVAL '1 MINUTE' AND NOW() " +
```
dhirajshettymoj, line 15: what if we remove this 1 minute check altogether even if they are not recent duplicates?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationService.kt
```diff
@@ -39,10 +41,23 @@ class VisitNotificationService(
     }
   }
 
+  private fun isNotADeleteEvent(nonAssociationChangedNotification: NonAssociationChangedNotificationDto): Boolean {
+    val nonAssociations = prisonerService.getOffenderNonAssociationList(nonAssociationChangedNotification.prisonerNumber)
```
dhirajshettymoj, line 45: what will happen if this call fails for timeout?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
```diff
@@ -1,8 +1,23 @@
...
+      " AND ve.visit_id=:visitId AND ve.type=:#{#nonAssociationEvent.name()}",
+    nativeQuery = true,
+  )
+  fun isEventARecentDuplicate(
+    visitId: Long,
+    nonAssociationEvent: NotificationEventType,
```
dhirajshettymoj, line 21: should var name be changed from nonAssociationEvent to notificationEvent?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationService.kt
dhirajshettymoj: good point. I think the getOffenderNonAssociationList will return the current non associations while the event could be for a future date. We will have to check the API docs.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
dhirajshettymoj: maybe this is needed as we can have a non association for a 3rd prisoner added.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationService.kt
AledWEvans: added extra checks

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationService.kt
AledWEvans: good point

dhirajshettymoj (approved):
LGTM

PR #534, opened 2023-09-22
Vb 1536 proactively manage bookings, added endpoints for new notifications

AledWEvans:
VB-1536 Proactively manage bookings, added endpoints for new notifications


dhirajshettymoj (approved):
LGTM

PR #538, opened 2023-09-26
Vb 2998 updating session template removes selected groups in admin service

AledWEvans:
VB-2998 Updating session template removes selected groups in admin service, fixes a bug


dhirajshettymoj (approved):
LGTM

PR #540, opened 2023-09-27
VB-1536 Proactively manage bookings, changed notification reference t…

AledWEvans:
VB-1536 Proactively manage bookings, changed notification reference to not null


dhirajshettymoj (approved):
LGTM

PR #553, opened 2023-10-19
VB-3039 Proactive bookings : Create server side endpoint to get infor…

AledWEvans:
Dummy end point to help UI guys get on with there dev


dhirajshettymoj (approved):
LGTM

PR #554, opened 2023-10-19
Vb 3039 proactive bookings create server side endpoint to get information back on the notification count, implementation

AledWEvans:
implementation of code to return the notification count


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
```diff
@@ -37,4 +37,24 @@ interface VisitNotificationEventRepository : JpaRepository<VisitNotificationEven
...
+
+  @Query(
+    "SELECT sum(ng) FROM (   " +
+      "SELECT count(distinct vne.reference) as ng FROM visit_notification_event vne " +
+      "   JOIN visit v ON v.reference  = vne.booking_reference " +
+      "   WHERE v.visit_start >= NOW() GROUP BY vne.reference) sq ",
+    nativeQuery = true,
+  )
+  fun getNotificationGroups(): Int
 }
```
AledWEvans, line 60: This logic above is a tricky one... not sure if it will work

dhirajshettymoj (approved):
LGTM

PR #557, opened 2023-10-23
VB-3054 Found a bug in the notification code

AledWEvans:
Found defect in notification code, the type was hard coded


dhirajshettymoj (approved):
LGTM

PR #559, opened 2023-10-24
VB-3054 Found bug with NA url, added correct URL, revert to old url, …

AledWEvans:
reverting code to old url


dhirajshettymoj (approved):
LGTM

PR #565, opened 2023-10-26
VB-2762 Incorrect hardcoded value in log needs to change to use passe…

AledWEvans:
fixed a small bug with logging text


dhirajshettymoj (approved):
LGTM

PR #566, opened 2023-10-26
Vb 2762 we were not including the start and end date

AledWEvans:
Start and end date for restriction, I was not including this originally because I've been told the dates are not used correctly in NOMIS, but I have changed my mind


dhirajshettymoj (approved):
LGTM

PR #568, opened 2023-11-02
VB-3172 - update visitsToDate to be null so that all visits will be returned

dhirajshettymoj:

Updates visits stats to not have an end date.

VB-3172 fix


AledWEvans (approved):
LGTM

PR #570, opened 2023-11-03
VB-3183 Migrate 2 visits from one session template to another

AledWEvans:
Migrate 2 visits from one session template to another


hutcheonb-moj (approved):
LGTM

PR #573, opened 2023-11-07
VB-2763 - update get visits to allow returning visits by session template reference and session date.

dhirajshettymoj:
…late reference and session date.

Update to get visits endpoint to add sessionTemplateReference and sessionDate parameters.

Fix for VB-2763.


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
```diff
@@ -335,6 +336,18 @@ class VisitController(
       example = "2021-11-03T09:00:00",
     )
     endDateTime: LocalDateTime?,
+    @RequestParam(value = "sessionTemplateReference", required = false)
```
dhirajshettymoj, line 339: should these be moved to a different endpoint instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
```diff
@@ -335,6 +336,18 @@ class VisitController(
...
+    @Parameter(
+      description = "Filter results by session template",
+      example = "v9-d7-ed-7u",
+    )
+    sessionTemplateReference: String?,
```
dhirajshettymoj, line 344: should these be moved to a different endpoint instead?

AledWEvans: I would say so because if you have the reference nothing else counts

PR #577, opened 2023-11-10
Vb 3192 create service side code in v si p service to group notifications by type

AledWEvans:
create service side code in v si p service to group notifications by type

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.visitscheduler.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
+import org.springframework.data.repository.query.Param
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.notification.VisitNotificationEvent
 import uk.gov.justice.digital.hmpps.visitscheduler.service.NotificationEventType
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationEventService.kt
```diff
@@ -91,15 +92,68 @@ class VisitNotificationEventService(
   }
 
   private fun processVisitsWithNotifications(affectedVisits: List<VisitDto>, type: NotificationEventType) {
+    val affectedVisitsNoDuplicate = affectedVisits.filter { !visitNotificationEventRepository.isEventARecentDuplicate(it.reference, type) }
+    flagTrackEvents(affectedVisitsNoDuplicate, type)
```
dhirajshettymoj, line 96: should we track this event only after we save to DB?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
```diff
@@ -21,6 +22,26 @@ interface VisitNotificationEventRepository : JpaRepository<VisitNotificationEven
...
+   * The inner query will only return one True or no Result because of the HAVING COUNT(reference)=2 and LIMIT 1 the outer query will then
+   * return true if a result otherwise false. If we did not have the outer query we would get a null pointer.
+   *
+   * This is a very efficient way of doing these calculations using code will be inefficient
+   */
+  @Query(
+    "SELECT count(*) = 1 from (SELECT COUNT(*) = 2 FROM visit_notification_event vne " +
+      " WHERE vne.create_timestamp BETWEEN NOW() - INTERVAL '10 MINUTE' AND NOW() AND " +
+      "   vne.booking_reference in (:bookingReferencePair1,:bookingReferencePair2) AND " +
+      "   vne.type=:#{#notificationEvent.name()} " +
```
dhirajshettymoj, line 35: what if the event is an non association update event and not a create event? Does the code need to differentiate between them?

AledWEvans: good point

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationEventService.kt
AledWEvans: good point,...

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/VisitNotificationEventRepository.kt
AledWEvans: if (NON_ASSOCIATION_CREATED == notificationDto.type) {
      val prisonCode = prisonerService.getPrisonerSupportedPrisonCode(notificationDto.prisonerNumber)
      prisonCode?.let {
        val affectedVisits = getOverLappingVisits(notificationDto, prisonCode)
        processVisitsWithNotifications(affectedVisits, NON_ASSOCIATION_EVENT)
      }
    } else if (notificationDto.type in arrayOf(NON_ASSOCIATION_DELETED, NON_ASSOCIATION_CLOSED)) {

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitNotificationEventService.kt
AledWEvans: After talk with @dhirajshettymoj decided to leave as is since this may be taken out and also report should consistent format no matter what the event type

dhirajshettymoj (approved):
LGTM

PR #578, opened 2023-11-13
VB-3197 Remove video call visits for AYI from database

AledWEvans:
As part of the migrations for onboarding AYI, we included some video call visits that the establishment had put into the wrong NOMIS room.

We need toe remove these from the database so they do not appear incorrectly as SOCIAL visits.


dhirajshettymoj (approved):
LGTM

PR #579, opened 2023-11-14
VB-3115 - set productId to DPS035

dhirajshettymoj:
VB-3115 -

- set productId to DPS035
- remove contextColour, v1_2_enabled and v0_47_enabled as these are no longer needed.
 

set productId to DPS035 

VB-3115

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -6,9 +6,6 @@ generic-service:
 
   ingress:
     host: visit-scheduler-dev.prison.service.justice.gov.uk
-    contextColour: green
-    v1_2_enabled: true
-    v0_47_enabled: false
 
   env:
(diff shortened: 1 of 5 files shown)


psoleckimoj (approved):
LGTM!

helm_deploy/values-dev.yaml
```diff
@@ -6,9 +6,6 @@ generic-service:
 
   ingress:
     host: visit-scheduler-dev.prison.service.justice.gov.uk
-    contextColour: green
```
dhirajshettymoj, line 9: removed as not used on the new projects created from template projects and also mentioned here that it is not needed for the new helm charts - 
https://mojdt.slack.com/archives/C04JFG3QJE6/p1698751082281769?thread_ts=1698749911.096719&cid=C04JFG3QJE6

PR #586, opened 2023-11-23
VB-2763 - add sql script to delete video call visits that are in Prod.

dhirajshettymoj:

Deletes visits that are less then 45 minutes from DB for Phase B and Phase C prisons.

Data migration clean up.


AledWEvans (approved):
LGTM

PR #602, opened 2023-12-20
VB-1845 Restrict visits migrated to n number of days

AledWEvans:
Currently when we migrate visits from NOMIS we bring across all visits for a prison. We have noticed that these visits are sometimes too far ahead and might have been created on NOMIS in error. 

As a fix, we need to set an upper limit on how far in the future we need to migrate visits. We have decided that it will be 6 months for now but this date will be configurable.

test were initially failing because of hard coded date a year in future, but simple fix


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
```diff
@@ -48,12 +51,18 @@ class MigrateVisitService(
...
+  @Value("\${migrate.max.in.future.months:6}")
+  private val migrateMaxInFutureMonths: Long,
 ) {
 
   @Autowired
   private lateinit var visitDtoBuilder: VisitDtoBuilder
 
   fun migrateVisit(migrateVisitRequest: MigrateVisitRequestDto): String {
+    if (isVisitTooFarInTheFuture(migrateVisitRequest.startTimestamp)) {
+      throw VisitToMigrateException("Visit more than $migrateMaxInFutureMonths month's in future, will not be migrated!")
```
dhirajshettymoj, line 63: minor typo - month's needs to be months.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/config/VisitSchedulerExceptionHandler.kt
```diff
@@ -87,6 +88,18 @@ class VisitSchedulerExceptionHandler(
     return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error)
   }
 
+  @ExceptionHandler(VisitToMigrateException::class)
```
dhirajshettymoj, line 91: Not very clear on the migration process - this exception when thrown will impact only 1 visit and the migration process will continue. Is that correct?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
```diff
@@ -148,6 +157,10 @@ class MigrateVisitService(
     return visitEntity.reference
   }
 
+  private fun isVisitTooFarInTheFuture(date: LocalDateTime): Boolean {
```
dhirajshettymoj, line 160: sorry minor again - visitDate?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/config/VisitSchedulerExceptionHandler.kt
AledWEvans: yes thats correct,... each visit is migrated one at a time,.. and associated info is given for syscon

dhirajshettymoj (dismissed):
LGTM

dhirajshettymoj (approved):

PR #607, opened 2024-01-04
Vb 3349 visit scheduler service, api changes and sql query to return notifications

AledWEvans:
Vb 3349 visit scheduler service, api changes and sql query to return notifications


dhirajshettymoj (approved):
LGTM

PR #608, opened 2024-01-05
VB-3358 - remove associated notification events with cancellation

dhirajshettymoj:

Removes any notification events associated with a visit reference from the visit notification event table.

VB-3358 requirements


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -406,9 +406,7 @@ class VisitService(
     }
   }
 
-  private fun validateCancelRequest(reference: String) {
-    val visitEntity =
```
dhirajshettymoj, line 410: this code was being duplicated hence removed from here.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -331,10 +333,10 @@ class VisitService(
       return visitDtoBuilder.build(canceledVisit)
     }
 
-    validateCancelRequest(reference)
+    val visitEntity = visitRepository.findBookedVisit(reference) ?: throw VisitNotFoundException("Visit $reference not found")
```
dhirajshettymoj, line 336: moved this before call to validate as this call was being duplicated.

AledWEvans (approved):
LGTM

PR #613, opened 2024-01-12
VB-3366,VB-3365,VB-3368  flyway script, migration of data code and DB…

AledWEvans:
Flyway scripts,
Which also contain migration code and implements DB structure


src/main/resources/db/migration/V3_00__create_application_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+            booking_reference text NULL,
+            visit_type varchar(80) NULL,
+            restriction varchar(80) NULL,
+            created_by varchar(60) NULL,
+            create_timestamp timestamp NULL,
+            modify_timestamp timestamp NULL
+);
+
+-- Create session slot table
+CREATE TABLE session_slot
```
dhirajshettymoj, line 36: will this be created for all dates when we create a session template?

AledWEvans: Narr this will be created in a lazy fashion when the first booking is booked for that slot,... after that it will just reuse for that slot.

PR #614, opened 2024-01-12
Bristol migration data scripts - created by Simon for migration of visits between Bristol sessions post timetable changes.

dhirajshettymoj:

Bristol migration scripts.

Bristol migration scripts to be run on Live to move visits between sessions due to timetable changes at Bristol


AledWEvans (approved):
LGTM

PR #617, opened 2024-01-12
Bug that tom and ben found, some tests were written incorrectly also.

AledWEvans:
Code was matching to visits with the same prisoner id


dhirajshettymoj (approved):
LGTM

dhirajshettymoj (commented):
test

dhirajshettymoj (approved):
LGTM

PR #622, opened 2024-01-25
VB-3447 fix - delete PRISON_VISITS_BLOCKED_FOR_DATE visit notification events when the visit date is updated

dhirajshettymoj:

Deletes any PRISON_VISITS_BLOCKED_FOR_DATE when the visit is updated and moved to a new date.

Fix issues

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -305,6 +305,8 @@ class VisitService(
       existingBooking.visitStatus = CANCELLED
       existingBooking.outcomeStatus = SUPERSEDED_CANCELLATION
       visitRepository.saveAndFlush(existingBooking)
+    }?.also { existingBooking ->
+      handleVisitUpdateEvents(existingBooking, visitToBook)
     }
 
     visitToBook.visitStatus = BOOKED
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -504,6 +506,12 @@ class VisitService(
     }
   }
 
+  private fun handleVisitUpdateEvents(existingBooking: Visit, updatedVisit: Visit) {
+    if (existingBooking.visitStart.toLocalDate() != updatedVisit.visitStart.toLocalDate()) {
+      visitNotificationEventRepository.deleteByBookingReferenceAndType(existingBooking.reference, NotificationEventType.PRISON_VISITS_BLOCKED_FOR_DATE)
+    }
```
AledWEvans, line 512: I suppose just your just doing PRISON_VISITS_BLOCKED_FOR_DATE for now

AledWEvans (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
dhirajshettymoj: yes the other ones do not change with a visit date change. we might have to revisit this for non associations.

PR #623, opened 2024-02-01
VB-3335    Separate visits booked from visit applications

AledWEvans:
Problem Statement

Currently on an implementation level we have the concept of a visit (god table) in the database.  

Visits only exiting when they are booked (or temporarily as a reservation that expire) - with new public service we may need the concept of an ‘application’ when visitors apply to visit someone.

This visit can have duplicates for the same application and time slot, on a software architecture level this causing complexity in our code and SQL queries. 

The solution to this is to apply data modelling / normalisation techniques to break the table into sub-tables that correspond to the actual real world use within reason. The idea behind this is that the code does not have add complexity to process data that belongs to multiple forms. 

The problem with the god table approach is that a god table often belongs to one or more actual processes therefore in order to comply with multiple business processes we are forced to: 

duplicate data

change the condition of the data for all process so that it compatible for one process

handle performance issues in the long run

handle migration issues

create code that is complex and less maintainable because it has to cater for the missing relationships in the data

introduce considerable complexity to deal with time slot availability information, this takes considerable time to process and slows down the user experience

create booking references before we have created a booking

create a visit before we have a visit

have SQL/Kotlin code that has to decipher the difference between an application reservation and a visit 

manage issues related to abandoned reserved bookings: see VB-2295: Investigate how we can better handle expired reservations
BACKLOG


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/EventAuditRepository.kt
```diff
@@ -32,11 +32,12 @@ interface EventAuditRepository : JpaRepository<EventAudit, Long> {
   @Transactional
   @Modifying
   @Query(
-    "Update event_audit SET application_method_type=:#{#applicationMethodType.name()}  " +
+    "Update event_audit SET application_method_type=:#{#applicationMethodType.name()},  " +
```
dhirajshettymoj, line 35: would this be an insert everytime instead of update?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/ApplicationService.kt
```diff
@@ -0,0 +1,333 @@
...
+  private fun saveEventAudit(
+    actionedBy: String,
+    application: ApplicationDto,
+    visit: Visit? = null,
+    applicationMethodType: ApplicationMethodType,
+  ) {
+    eventAuditRepository.saveAndFlush(
+      EventAudit(
+        actionedBy = actionedBy,
+        bookingReference = visit?.let { visit.reference } ?: "",
```
dhirajshettymoj, line 260: null instead of ""?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/EventAuditRepository.kt
AledWEvans: narr this is to update the application method type at the end of the process.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/ApplicationService.kt
AledWEvans: not sure could be null... it's probably better not to include it, I will do that

dhirajshettymoj (approved):
Approved after initial testing - merging to develop branch. Plan is to do some further tests on dev and then merge it to main post more checks.

PR #624, opened 2024-02-02
VB-3335 Defect found slot sql, duplicate session slot were being prod…

AledWEvans:
Duplicate session slot's were being created in preprod,.. have fixed this by changing the group by


dhirajshettymoj (approved):
LGTM

PR #627, opened 2024-02-06
added a CreateApplicationRestriction, as these options should be diff…

AledWEvans:
added a CreateApplicationRestriction, as these options should be different from the server side model


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/application/CreateApplicationRestriction.kt
```diff
@@ -0,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.dto.application
+
+import uk.gov.justice.digital.hmpps.visitscheduler.model.VisitRestriction
+import java.lang.IllegalArgumentException
+
+@Suppress("unused")
+enum class CreateApplicationRestriction(
+  val description: String,
```
dhirajshettymoj, line 8: we could have the VisitRestrition as a parameter here but works as it is currently as well

dhirajshettymoj (approved):
LGTM - 1 minor comment to consider but can be ignored.

PR #629, opened 2024-02-07
Changes to fix updating ok booking,.. results were not being saved

AledWEvans:
Changes to fix updating ok booking,.. results were not being saved now they are

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -108,7 +108,7 @@ class VisitService(
     return bookedVisitDto
   }
 
-  private fun createBooking(application: Application, hasExistingBooking: Boolean): Visit {
+  fun createBooking(application: Application, hasExistingBooking: Boolean): Visit {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -146,12 +146,23 @@ class VisitService(
...
+          visitId = booking.id,
+          name = it.name,
+          telephone = it.telephone,
+        )
+      }
     }
 
     application.support.let {
       booking.support.clear()
+      visitRepository.saveAndFlush(booking)
```
dhirajshettymoj, line 165: are we doing  save here but not in line 150 because we clear it out?

dhirajshettymoj (approved):
LGTM - 1 minor query added

tpmcgowan (commented):
Have tested out locally and it works - will leave approving the code to Dhiraj 😀

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
AledWEvans: yes

PR #637, opened 2024-02-16
Add indexes to visit and application as joins with session_slot are r…

dhirajshettymoj:

Adds index to make joins between application and session_slot and visit and session_slot a bit quicker.

Performance improvements


tpmcgowan (approved):
LGTM

PR #640, opened 2024-02-22
VB-3545 Improved migration script to increase speed

AledWEvans:
While deploying to preprod we had some speed issues.

So have split up sql's into separate files to hopefully have separate time outs per file.
Also added an sql to improve update of application reserved false on visit migrations to application,.. prior to this it was almost always true


src/main/resources/db/migration/V3_04__insert_data_into_application_no_order_table.sql
```diff
@@ -0,0 +1,41 @@
+CREATE TEMP TABLE tmp_not_reserved_applications(application_reference text UNIQUE NOT NULL);
+
+-- create temp table to store not reserved applications
+INSERT INTO tmp_not_reserved_applications(application_reference)
```
dhirajshettymoj, line 4: this might not be needed if done during after hours.

dhirajshettymoj (approved):
LGTM

src/main/resources/db/migration/V3_04__insert_data_into_application_no_order_table.sql
AledWEvans: its used for historic data not current data

PR #646, opened 2024-02-26
Moved livenessProbe and readinessProbe in prod and preprod only

AledWEvans:
Problems with dev and staging, these settings are not required here.


dhirajshettymoj (approved):
LGTM

PR #652, opened 2024-03-01
VB-3568 Some minor performance issues in production when selecting ti…

AledWEvans:
used session slot id to get visits or applications, also added indexes


dhirajshettymoj (approved):
LGTM

PR #653, opened 2024-03-04
VB-3413 - Allow optional contact

dhirajshettymoj:

Allows contact to be optional

New requirements


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/TelemetryClientService.kt
```diff
@@ -75,6 +75,7 @@ class TelemetryClientService(
       "visitStart" to formatDateTimeToString(visit.startTimestamp),
       "visitStatus" to visit.visitStatus.name,
       "applicationReference" to visit.applicationReference,
+      "hasContactInformation" to ((visit.visitContact != null).toString()),
```
dhirajshettymoj, line 78: adding this for all scenarios - booked / updated / cancelled in case we want to count how many messages are sent

dhirajshettymoj:
@AledWEvans have updated the code to include migration scenarios as suggested. Thanks for the spot

src/main/resources/db/migration/V3_17__update_application_and_visit_contact_number_to_null_where_unknown.sql
```diff
@@ -0,0 +1,2 @@
+update application_contact set contact_phone = NULL where upper(contact_phone) = 'UNKNOWN';
+update visit_contact set contact_phone = NULL where upper(contact_phone) = 'UNKNOWN';
```
AledWEvans, line 2: Why uppper should always be upper case?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/TelemetryClientService.kt
```diff
@@ -85,6 +85,7 @@ class TelemetryClientService(
       "visitStart" to formatDateTimeToString(visit.startTimestamp),
       "visitStatus" to visit.visitStatus.name,
       "applicationReference" to visit.applicationReference,
+      "hasPhoneNumber" to ((visit.visitContact.telephone != null).toString()),
```
AledWEvans, line 88: what happens when you do just
visit.visitContact.telephone?

AledWEvans (approved):
LGTM, some comments your call!

src/main/resources/db/migration/V3_17__update_application_and_visit_contact_number_to_null_where_unknown.sql
dhirajshettymoj: will leave it. just an additional check.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/TelemetryClientService.kt
dhirajshettymoj: Need the true or false value as a string as we return value as MutableMap<String, String>.

PR #657, opened 2024-03-05
VB-3568 Some minor performance issues in production when selecting ti…

AledWEvans:
Rewrote code to create visit sessions as performance was not great


dhirajshettymoj (approved):
LGTM. reduces number of queries and also improves code flow. We should see some major performance improvements!

PR #659, opened 2024-03-07
VB-3585 adding ability to delete applications and visits via the test…

AledWEvans:
in correct swagger doc details


dhirajshettymoj (approved):
LGTM

PR #661, opened 2024-03-12
VB-3621 Migration bug - found with duplicate data on session slots

AledWEvans:

add's constraint and try catch code to return the correct value if issue occurs

There is a race condition and in those cases a extra entry is entered into the DB, and all other visits for that slot then fail because the JPA query tries to return multiple results when only one is permitted.


dhirajshettymoj (approved):
LGTM

PR #668, opened 2024-03-19
Update visit slots and session slots for newly created session templates for Bullingdon.

dhirajshettymoj:

Updates session slot Ids for affected visits.

Manual data correction


simonnebMOJ (approved):
LTGM

PR #677, opened 2024-04-15
VB-3723 Add age config to VSIP prison API and database,

AledWEvans:
removed -1 option on child visitors, also removed haed coded check for max number of visitors.


dhirajshettymoj (approved):
LGTM

PR #684, opened 2024-04-25
VB-3719 and VB-3718 to allow for PUBLIC and STAFF supported prisons

AledWEvans:
Allow for PUBLIC and STAFF supported prisons


tpmcgowan (commented):
Haven't reviewed the code but have tested locally with Postman and looks good 👍

PR #686, opened 2024-04-25
Vb 3769 alter application api to cater for public and staff sources

AledWEvans:
Added UserType


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/ApplicationService.kt
```diff
@@ -117,6 +117,7 @@ class ApplicationService(
         reservedSlot = isReservedSlot,
         visitType = sessionTemplate.visitType,
         restriction = createApplicationDto.applicationRestriction.getVisitRestriction(),
+        userType = createApplicationDto.userType,
```
dhirajshettymoj, line 120: should this go into App Insights as well?

dhirajshettymoj, line 120: also does eventAudit need to change as a result. Can be a a part of a different story.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/application/CreateApplicationDto.kt
```diff
@@ -35,4 +36,7 @@ data class CreateApplicationDto(
   var visitorSupport: ApplicationSupportDto? = null,
   @Schema(description = "Username for user who actioned this request", required = true)
```
dhirajshettymoj, line 37: will this need to change based on userType?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/ApplicationService.kt
AledWEvans: different ticket

AledWEvans: This needs to be done,.. but not for now

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/application/CreateApplicationDto.kt
AledWEvans: The actionedBy will use the booker reference sent down from font end,... the orchestration now deals with it..

dhirajshettymoj (approved):

tpmcgowan (approved):
LGTM

PR #687, opened 2024-04-26
Vb 3698 create new session vsip endpoint add session type

dhirajshettymoj:

Replace VisitRestriction with SesssionType as parameter for GET available sessions.

VB-3698


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/SessionType.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.model
+
+enum class SessionType(
```
dhirajshettymoj, line 3: named enum as SessionType instead on SessionRestriction as it might be confused with VisitRestriction.

AledWEvans (approved):
LGTM

PR #691, opened 2024-05-02
Vb 3800 add to available sessions to check prisoner has closed restrictions

AledWEvans:
add to available sessions to check prisoner has closed restrictions,

Also changes api to use dates instead of min max on VSiP


dhirajshettymoj (approved):
LGTM

PR #693, opened 2024-05-09
Update DB Script to source change to visit.

AledWEvans:
added source or user type to visit and migration


dhirajshettymoj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/Visit.kt
```diff
@@ -61,6 +62,11 @@ class Visit(
   @Column(nullable = false)
   @Enumerated(EnumType.STRING)
   var visitRestriction: VisitRestriction,
+
+  @Enumerated(EnumType.STRING)
```
adaviesMOJ, line 66: do we need the 

@Column(nullable = false) annotation? As this seems like it will now be a required field?

AledWEvans: yes should follow the DB column....

adaviesMOJ (approved):

PR #694, opened 2024-05-10
Update config and README to be in line with new local development

adaviesMOJ:

This PR will remove any outdated docker-compose files and application-<>.yml files to only leave the required configurations to run the service locally. Also update README to document process.


AledWEvans (approved):
LGTM

PR #695, opened 2024-05-13
VB-3838 Update BFI migrated visits to correct start and end time

AledWEvans:

Fixes up migrated data in DB

To make sure booking can be altered after migration see VB-3838


dhirajshettymoj (approved):
LGTM

PR #696, opened 2024-05-14
VB-3839 Update migration script so that session template start and en…

AledWEvans:
During migration of visits from NOMIS to VSIP, the start and end time of the sessions in NOMIS are preserved. The visits themselves are associated with the nearest matching visit session.

Sometimes, these visit start and end times do not match (for example NOMIS might have been 14:15 - 15:15 and VSIP has 14:00 - 15:00.

The service will think that the prisoner is already on another visit, and will not allow the visit  to be updated. 

In order to fix this we need to update the migration code so that the visit times of the migrated visits match the visit session they are assigned to.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.visitscheduler.service
 
 import com.microsoft.applicationinsights.TelemetryClient
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
```diff
@@ -58,6 +60,10 @@ class MigrateVisitService(
   private val migrateMaxMonthsInFuture: Long,
 ) {
 
+  companion object {
+    val LOG: Logger = LoggerFactory.getLogger(this::class.java)
```
adaviesMOJ, line 64: tiny point but should this be called log in lowercase?

dhirajshettymoj (approved):
LGTM

adaviesMOJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/MigrateVisitService.kt
AledWEvans: Well I see this as constant and constant should be in upper case.

PR #724, opened 2024-05-30
VB-3915 Store booker reference, new action_by table and migration

AledWEvans:
Adds new way of storing the booker reference.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/audit/ActionedByDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/audit/ActionedByDto.kt
@@ -1,0 +1,26 @@
+package uk.gov.justice.digital.hmpps.visitscheduler.dto.audit
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotNull
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.UserType
+import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.ActionedBy
+
+@Schema(description = "Actioned By")
+class ActionedByDto(
+
+  @Schema(description = "booker reference", example = "asd-aed-vhj", required = true)
+  val bookerReference: String?,
+
+  @Schema(description = "User Name", example = "AS/ALED", required = false)
+  val userName: String?,
+
+  @Schema(description = "User type", example = "STAFF", required = false)
+  @field:NotNull
+  val userType: UserType,
+) {
+  constructor(entity: ActionedBy) : this(
+    bookerReference = entity.bookerReference,
+    userName = entity.userName,
+    userType = entity.userType,
+  )
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/audit/ActionedByDto.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/audit/ActionedByDto.kt
@@ -8,7 +8,7 @@ import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.ActionedBy
 @Schema(description = "Actioned By")
 class ActionedByDto(
 
-  @Schema(description = "booker reference", example = "asd-aed-vhj", required = true)
+  @Schema(description = "booker reference", example = "asd-aed-vhj", required = false)
   val bookerReference: String?,
 
   @Schema(description = "User Name", example = "AS/ALED", required = false)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
@@ -1,7 +1,5 @@
 package uk.gov.justice.digital.hmpps.visitscheduler.model.entity
 
-import jakarta.persistence.CascadeType.MERGE
-import jakarta.persistence.CascadeType.PERSIST
 import jakarta.persistence.CascadeType.REFRESH
 import jakarta.persistence.Column
 import jakarta.persistence.Entity
@@ -50,7 +48,7 @@ class EventAudit private constructor(
   @Column(name = "ACTIONED_BY_ID", nullable = true)
   private val actionedById: Long? = null,
 
-  @ManyToOne(fetch = FetchType.LAZY, cascade = [REFRESH, PERSIST, MERGE])
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [REFRESH])
   @JoinColumn(name = "ACTIONED_BY_ID", updatable = false, insertable = false)
   val actionedBy: ActionedBy,
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
@@ -45,8 +45,8 @@ class EventAudit private constructor(
   @Enumerated(EnumType.STRING)
   val applicationMethodType: ApplicationMethodType,
 
-  @Column(name = "ACTIONED_BY_ID", nullable = true)
-  private val actionedById: Long? = null,
+  @Column(name = "ACTIONED_BY_ID", nullable = false)
+  private val actionedById: Long,
 
   @ManyToOne(fetch = FetchType.LAZY, cascade = [REFRESH])
   @JoinColumn(name = "ACTIONED_BY_ID", updatable = false, insertable = false)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
@@ -1,24 +1,29 @@
 package uk.gov.justice.digital.hmpps.visitscheduler.model.entity
 
+import jakarta.persistence.CascadeType.MERGE
+import jakarta.persistence.CascadeType.PERSIST
+import jakarta.persistence.CascadeType.REFRESH
 import jakarta.persistence.Column
 import jakarta.persistence.Entity
 import jakarta.persistence.EnumType
 import jakarta.persistence.Enumerated
+import jakarta.persistence.FetchType
 import jakarta.persistence.GeneratedValue
 import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
 import jakarta.persistence.Table
 import org.hibernate.annotations.CreationTimestamp
 import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.ApplicationMethodType
 import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.EventAuditType
-import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.UserType
 import java.time.LocalDateTime
 
 @Entity
 @Table(
   name = "event_audit",
 )
-class EventAudit(
+class EventAudit private constructor(
 
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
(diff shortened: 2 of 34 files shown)


src/main/resources/db/migration/V3_26__add_new_actioned_by_table.sql
```diff
@@ -0,0 +1,121 @@
...
+        user_type               VARCHAR(80)     NOT NULL,
+        create_timestamp        timestamp       NOT NULL default current_timestamp
+    );
+
+    CREATE INDEX idx_action_by_user_type ON actioned_by(user_type);
+
+    INSERT INTO actioned_by(user_name,user_type)
+            SELECT actioned_by,user_type FROM event_audit
+                                         WHERE "type" not in ('PRISON_VISITS_BLOCKED_FOR_DATE','PRISONER_RELEASED_EVENT')
+                                         group by actioned_by,user_type;
```
AledWEvans, line 15: Should I bother doing PUBLIC userType here also?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/EventAuditRepository.kt
```diff
@@ -22,12 +23,11 @@ interface EventAuditRepository : JpaRepository<EventAudit, Long> {
   fun findLastBookedVisitEventByBookingReference(bookingReference: String): EventAudit?
 
   @Query(
-    "SELECT actioned_by FROM event_audit " +
-      "WHERE booking_reference = :bookingReference AND type in ('UPDATED_VISIT','BOOKED_VISIT') " +
-      "ORDER BY id DESC LIMIT 1 ",
-    nativeQuery = true,
+    "SELECT ea.actionedBy FROM EventAudit ea " +
```
adaviesMOJ, line 26: Question: why did you change this from native SQL to JPQL?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitEventAuditService.kt
```diff
@@ -0,0 +1,212 @@
...
+
+  @Lazy
+  @Autowired
+  private lateinit var actionedByRepository: ActionedByRepository
+
+  @Lazy
+  @Autowired
+  private lateinit var eventAuditRepository: EventAuditRepository
+
+  fun saveEventAudit(
```
adaviesMOJ, line 55: Suggestion: Is it worth giving these methods their own names? As they technically save "different" eventAudits?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitEventAuditService.kt
```diff
@@ -0,0 +1,212 @@
...
+
+  @Transactional(readOnly = true)
+  fun getLastEventForBooking(bookingReference: String): EventAuditDto? {
+    return eventAuditRepository.findLastBookedVisitEventByBookingReference(bookingReference)?.let {
+      EventAuditDto(it)
+    }
+  }
+
+  @Transactional(readOnly = true)
+  fun getLastUserToUpdateToSlotByReference(bookingReference: String): String {
```
adaviesMOJ, line 175: rename to: getLastUserToUpdateSlotByReference ?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
```diff
@@ -1,58 +1,74 @@
...
 
-  @Column(nullable = false)
-  var actionedBy: String,
+  @Column(name = "ACTIONED_BY_ID", nullable = true)
+  private val actionedById: Long? = null,
 
-  @Enumerated(EnumType.STRING)
-  @Column(name = "user_type", nullable = false)
-  val userType: UserType,
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [REFRESH, PERSIST, MERGE])
```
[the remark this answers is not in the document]
AledWEvans: probably don't,... refresh makes sense...

src/main/resources/db/migration/V3_26__add_new_actioned_by_table.sql
```diff
@@ -0,0 +1,121 @@
...
+
+
+    -- For Notification as they do not have a source
+    INSERT INTO actioned_by(user_type) VALUES ('SYSTEM');
```
adaviesMOJ, line 19: Question: How does this insert statement work? There is no select, it just seems to be inserting directly with no condition?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/repository/EventAuditRepository.kt
AledWEvans: it was an easy way to return the actionedBy with out having to do a join

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitEventAuditService.kt
AledWEvans: I did consider this was in two minds,... so decided to go with the overload approach.

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitTest.kt
```diff
@@ -647,10 +682,18 @@ class BookVisitTest : IntegrationTestBase() {
...
 
-  private fun assertAuditEvent(visitDto: VisitDto, visitEntity: Visit) {
+  private fun assertAuditEvent(visitDto: VisitDto, visitEntity: Visit, userType: UserType = STAFF) {
     val eventAudit = this.eventAuditRepository.findLastEventByBookingReference(visitDto.reference)
     assertThat(eventAudit.type).isEqualTo(EventAuditType.BOOKED_VISIT)
-    assertThat(eventAudit.actionedBy).isEqualTo("booking_guy")
+    assertThat(eventAudit.actionedBy).isNotNull()
+    assertThat(eventAudit.actionedBy.userType).isEqualTo(userType)
+    if (STAFF == eventAudit.actionedBy.userType) {
+      assertThat(eventAudit.actionedBy.userName).isEqualTo("booking_guy")
```
adaviesMOJ, line 691: could assert that the other field is null ? To make sure we're only inserting into the 1 column per user type?

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitTest.kt
```diff
@@ -647,10 +682,18 @@ class BookVisitTest : IntegrationTestBase() {
...
+    }
+    if (PUBLIC == eventAudit.actionedBy.userType) {
+      assertThat(eventAudit.actionedBy.bookerReference).isEqualTo("booking_guy")
```
adaviesMOJ, line 694: same here

could assert that the other field is null ? To make sure we're only inserting into the 1 column per user type?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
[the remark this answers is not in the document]
adaviesMOJ: what does refresh do? How would that column be "refreshed" ?

src/main/resources/db/migration/V3_26__add_new_actioned_by_table.sql
AledWEvans: Thats exactly what its doing, it connects the data later on in line 75

adaviesMOJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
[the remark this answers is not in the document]
AledWEvans: I means you can refresh the object from the DB using the refresh method

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/audit/ActionedByDto.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotNull
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.UserType
+import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.ActionedBy
+
+@Schema(description = "Actioned By")
+class ActionedByDto(
+
+  @Schema(description = "booker reference", example = "asd-aed-vhj", required = true)
```
dhirajshettymoj, line 11: required - false?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/TelemetryClientService.kt
```diff
@@ -151,7 +151,9 @@ class TelemetryClientService(
 
     bookingEventAudit?.let {
       flagEventDataMap["visitBooked"] = formatDateTimeToString(it.createTimestamp)
-      flagEventDataMap["actionedBy"] = it.actionedBy
+      flagEventDataMap["actionedBy"] = bookingEventAudit.actionedBy.userName ?: ""
+      flagEventDataMap["bookerReference"] = bookingEventAudit.actionedBy.bookerReference ?: ""
```
dhirajshettymoj, line 155: we could not send these values to appinsights if not null, but nothing defined on the story as such

dhirajshettymoj (commented):
2 minor comments

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
```diff
@@ -1,58 +1,72 @@
...
 
   @Column(nullable = false)
   @Enumerated(EnumType.STRING)
-  var applicationMethodType: ApplicationMethodType,
+  val applicationMethodType: ApplicationMethodType,
 
-  @Column(nullable = false)
-  var actionedBy: String,
+  @Column(name = "ACTIONED_BY_ID", nullable = true)
+  private val actionedById: Long? = null,
```
adaviesMOJ, line 49: If the actionedBy is a SYSTEM_USER, would it not have an actionedById?

adaviesMOJ: Basically, does this need to be nullable?

src/main/resources/db/migration/V3_26__add_new_actioned_by_table.sql
adaviesMOJ: If this goes in before public booking goes live, it should be fine?

dhirajshettymoj (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/helper/EntityDataLoader.kt
```diff
@@ -409,11 +414,12 @@ class VisitEntityHelper(
 @Transactional
 class EventAuditEntityHelper(
   private val eventAuditRepository: TestEventAuditRepository,
+  private val testActionedByRepository: TestActionedByRepository,
```
adaviesMOJ, line 417: Why do we have Test repositories? Why don't we use the actual repositories like we do in other Integration tests?

src/main/resources/db/migration/V3_26__add_new_actioned_by_table.sql
AledWEvans: should go live soon,...

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/model/entity/EventAudit.kt
AledWEvans: needs to be nullable = false,.... I think?

adaviesMOJ: Yes I think so.
PUBLIC / STAFF / SYSTEM, all of these will contain an Actioned_by_id, with different columns within the ActionedBy table being null...

But the ID field is mandatory right?

adaviesMOJ (approved):

PR #729, opened 2024-06-05
Fix: Stop non-association client bringing back other prisons

adaviesMOJ:

We only want to check non associations for prisoners at the same prison. Change flag to false to stop including other prisons.


adaviesMOJ:
We chose to avoid updating the endpoint for now, as we don't want to push this live without extensive testing to confirm the new endpoint behaves the same.

Closing.

PR #732, opened 2024-06-05
Fix: Introduce a gate way timeout exception handler

adaviesMOJ:

Aim of this change is to get a more specific response code than the generic 500 that the spring web client returns.


adaviesMOJ:
Closing this PR for now. It's unclear of how the slack alerts will function until tested and tweaked over time. This may eventually be required, but closing until then.

PR #734, opened 2024-06-07
VB-3956 Exclude application reference from session window

AledWEvans:
I have pushed this up into the feature by mistake.

Please do not merge


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
```diff
@@ -142,8 +142,14 @@ class VisitSessionController(
       example = "2024-12-20",
     )
     toDate: LocalDate,
+    @RequestParam(value = "excludeCurrentApplicationReference", required = false)
```
adaviesMOJ, line 145: name suggests its a boolean, could it be something like "excludedApplicationReference" ?

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/enums/SessionConflict.kt
```diff
@@ -3,5 +3,5 @@ package uk.gov.justice.digital.hmpps.visitscheduler.dto.enums
 @Suppress("unused")
 enum class SessionConflict {
   NON_ASSOCIATION,
-  DOUBLE_BOOKED,
+  DOUBLE_RESERVATION,
```
adaviesMOJ, line 6: Question: will this require a frontend change, do they implement this enum?

dhirajshettymoj: will need updating on the Orchestration Service.

dhirajshettymoj (approved):
nothing other than @adaviesMOJ's comments.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/enums/SessionConflict.kt
AledWEvans: DOUBLE_BOOKING_OR_RESERVATION

dhirajshettymoj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/enums/SessionConflict.kt
```diff
@@ -3,5 +3,5 @@ package uk.gov.justice.digital.hmpps.visitscheduler.dto.enums
 @Suppress("unused")
 enum class SessionConflict {
   NON_ASSOCIATION,
-  DOUBLE_BOOKED,
+  DOUBLE_BOOKING_OR_RESERVATION,
```
tpmcgowan, line 6: If we're renaming this (and I do think the existing name is misleading), then does `EXISTING_BOOKING_OR_RESERVATION` better describe it? For me 'double' implies that we've **already** got two there - but isn't it actually that there is **one** there and it's warning you of the risk of a second?

PR #740, opened 2024-06-10
add logging for prisoner received transferred events

adaviesMOJ:
Add more output to logging debug an issue where lastPrisonId seems to be null...


adaviesMOJ:
No longer required as this method is being reworked to remove call to get prisoner details.

PR #743, opened 2024-06-11
VB-3735 Record which service made the booking (public or staff)

AledWEvans:
A refactor was needed with this code, so a bigger change than expected

Also Audit Event had to be done before the telemetry code, so those details could be stored.

The rework basically was to refactor code to conform to coding and architectural conventions.
Most of the code is just lift and shift from random place to the correct code service.

Also some of the namings were based on the old visit system (before application)


adaviesMOJ (approved):
LGTM -- Could the PR get a small description of what was changed and why ? As this includes a rework, it would be good to capture the basic thought process of why the rework was needed?

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/task/CleanUpApplicationsScheduleTest.kt
```diff
@@ -15,7 +15,7 @@ import org.springframework.boot.test.mock.mockito.MockBean
 import org.springframework.boot.test.mock.mockito.SpyBean
 import org.springframework.transaction.annotation.Propagation.SUPPORTS
 import org.springframework.transaction.annotation.Transactional
-import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.TelemetryVisitEvents.VISIT_SLOT_RELEASED_EVENT
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.TelemetryVisitEvents.APPLICATION_DELETED_EVENT
```
dhirajshettymoj, line 18: Simon might need to be informed of these changes  as his reports wil lbe impacted.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/enums/TelemetryVisitEvents.kt
```diff
@@ -4,9 +4,9 @@ enum class TelemetryVisitEvents(val eventName: String) {
   VISIT_SLOT_RESERVED_EVENT("visit-slot-reserved"),
   VISIT_CHANGED_EVENT("visit-changed"),
   VISIT_BOOKED_EVENT("visit-booked"),
-  VISIT_SLOT_CHANGED_EVENT("visit-slot-changed"),
+  APPLICATION_SLOT_CHANGED_EVENT("application-slot-changed"),
```
dhirajshettymoj, line 7: doubt it but maybe check with Simon if this is being used for any of his reports.

dhirajshettymoj (approved):
LGTM - needs some thorough regression by the team before release.

PR #745, opened 2024-06-14
Feature vb 3344 booking confirmed

AledWEvans:
Merge of Main into feature branch, had some old tests not working correctly so have also fixed these, also notices some failers with time checks when the expected time is out by a second


dhirajshettymoj (approved):
LGTM

PR #767, opened 2024-07-08
VB-4109 Return replicaCount back to 4 for prod

AledWEvans:
After big data migration we need to change replicaCount back to 4 and remove bespoke time out


dhirajshettymoj (approved):
LGTM

PR #772, opened 2024-07-12
Update error messages with error codes

dhirajshettymoj:

Replace validation exception with error codes

Backend validations.


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/ApplicationValidationService.kt
```diff
@@ -37,63 +47,90 @@ class ApplicationValidationService(
...
     // check capacity for slot
-    checkSlotCapacity(bookingRequestDto, application, existingBooking)
+    checkSlotCapacity(bookingRequestDto, application, existingBooking)?.also {
+      errorCodes.add(it)
+    }
+
+    return errorCodes
+  }
+
+  private fun isSystemApplicationValid(): List<ApplicationValidationErrorCodes> {
```
AledWEvans, line 123: not sure about using the is start to a method as this makes me think it will return a boolean
```
isStaffApplicationValid
isSystemApplicationValid
isPublicApplicationValid
```
May be use getStaffValidationErrors

AledWEvans (changes requested):
some naming changes would be good

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/ApplicationValidationService.kt
dhirajshettymoj: updated

adaviesMOJ (approved):
Nice mate!

PR #777, opened 2024-07-15
Update error response

dhirajshettymoj:

Update validation error response to return error code 422 instead of error code 400

Code review

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitValidationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitValidationTest.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.visitscheduler.integration.visit
 
+import io.netty.handler.codec.http.HttpResponseStatus.UNPROCESSABLE_ENTITY
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.BeforeEach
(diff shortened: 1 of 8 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitValidationTest.kt
```diff
@@ -123,11 +123,11 @@ class BookVisitValidationTest : IntegrationTestBase() {
     val responseSpec = callVisitBook(webTestClient, roleVisitSchedulerHttpHeaders, application.reference)
 
     // Then
-    responseSpec.expectStatus().isBadRequest
+    responseSpec.expectStatus().isEqualTo(422)
```
adaviesMOJ, line 126: Small suggestion which you can ignore if you want. Instead of

isEqualTo(422)

isEqualTo(UNPROCESSABLE_ENTITY.value()) to make it clear what the 422 is at a glance.

Approving anyway as its a preference.

adaviesMOJ (approved):
Approved with 1 comment, if you do change I'll review again 👍

adaviesMOJ (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/visit/BookVisitValidationTest.kt
dhirajshettymoj: Updated

PR #783, opened 2024-07-24
Changes for VB-4202 - send username as a parameter to getVisitSessions

dhirajshettymoj:
Changes for VB-4202 - send username as a parameter to getVisitSessions so that reserved applications by the STAFF user for the same prisoner are ignored.

Changes for VB-4202 - send username as a parameter to getVisitSession so that reserved applications by the STAFF user for the same prisoner are ignored.

VB-4202

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
@@ -85,8 +85,13 @@ class VisitSessionController(
       example = "28",
     )
     max: Int?,
+    @Parameter(
+      description = "Username for the user making the request. Optional, ignored if not passed in.",
+      example = "user-1",
+    )
+    username: String? = null,
   ): List<VisitSessionDto> {
-    return sessionService.getVisitSessions(prisonCode, prisonerId, minOverride = min, maxOverride = max)
+    return sessionService.getVisitSessions(prisonCode, prisonerId, minOverride = min, maxOverride = max, usernameToExcludeFromReservedApplications = username)
   }
 
   @PreAuthorize("hasRole('VISIT_SCHEDULER')")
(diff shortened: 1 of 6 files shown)


dhirajshettymoj:
Still uses createdBy on application. Picked up VB-4231 that will instead go to the event_audit table to find who created the application.

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
```diff
@@ -85,8 +85,13 @@ class VisitSessionController(
       example = "28",
     )
     max: Int?,
+    @Parameter(
+      description = "Username for the user making the request. Optional, ignored if not passed in.",
```
adaviesMOJ, line 89: Can we briefly explain what passing it in does?

E.g.

description = "Username for the user making the request. Used to exclude users pending applications from session capacity count. Optional, ignored if not passed in."

src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/session/GetSessionsDoubleBookingTest.kt
```diff
@@ -0,0 +1,169 @@
...
+import org.springframework.test.web.reactive.server.WebTestClient.BodyContentSpec
+import org.springframework.test.web.reactive.server.WebTestClient.ResponseSpec
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.SessionConflict
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.enums.VisitStatus
+import uk.gov.justice.digital.hmpps.visitscheduler.dto.sessions.VisitSessionDto
+import uk.gov.justice.digital.hmpps.visitscheduler.integration.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.visitscheduler.model.entity.session.SessionTemplate
+import java.time.LocalDate
+
+@DisplayName("Get /visit-sessions")
```
adaviesMOJ, line 16: This display name is being shared by 3 other test classes. Could we change them to be more unique?

E.g.

@DisplayName("Get /visit-sessions - Double Booking tests")

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitSessionController.kt
dhirajshettymoj: Thanks, good suggestion! Updated

adaviesMOJ (approved):

PR #785, opened 2024-07-25
Vb 4290

simonnebMOJ:

Moves visits associated with one session to another for Highpoint (HPI) - the start and end times were incorrect.

Ensuring the correct visits information is shown


dhirajshettymoj (approved):
LGTM

PR #790, opened 2024-07-31
push initial poc of single table strategy for event flagging

adaviesMOJ:

_Required._

_Required._


adaviesMOJ:
After discussions internally, we've agreed that no additional data capture is required. Instead of making this intelligent and un-flagging per flag, we're going to un-flag per visit which for now will handle everything we need.

JIRA link for new approach: https://dsdmoj.atlassian.net/browse/VB-4349

PR #793, opened 2024-08-01
VB-2961 - Enable processing of global visitor restrictions

adaviesMOJ:

Flag visits across all prisons for a visitor when they are given a new restriction that is supported by the visits service.


dhirajshettymoj (approved):
LGTM

PR #796, opened 2024-08-05
VB-4121 - Move from Prisoner API to Alerts API for prisoner alerts

adaviesMOJ:

Update the visit scheduler to remove any calls to retrieve alerts as the orchestration service now handles this. This should remove complexity around processing the alerts-updated event.


psoleckimoj (approved):
LGTM!

PR #798, opened 2024-08-06
VB-4274 - Fix connection reset issue

adaviesMOJ:

By introducing a set of max times for connections it should stop connections timing out while idle and then being used when they're not active.

Will monitor in dev for a while and see if rate of connection reset error reduces.

## JIRA Link
https://dsdmoj.atlassian.net/browse/VB-4274


psoleckimoj (approved):
LGTM!

PR #801, opened 2024-08-09
VB-4368 - Move event publishing outside of transaction - Part 1

adaviesMOJ:

Initial commit which contains a needed rework of some services to start splitting them into their own concerns. By untangling these services we can begin working to separate the event publishing out of the transaction for creating/updating/cancelling visits


hutcheonb-moj (approved):
LGTM

PR #802, opened 2024-08-12
VB-4368 - Move event publishing outside of transaction

adaviesMOJ:

Create a new service which handles storing and updating visit data, this allows us to remove the transaction at the top level function where the events are called to be processed, now outside of any transaction

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -4,7 +4,6 @@ import jakarta.validation.ValidationException
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Lazy
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.PageRequest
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -63,44 +49,40 @@ class VisitService(
...
 
-    val existingBooking = visitRepository.findVisitByApplicationReference(application.reference)
+    val booking = visitStoreService.createOrUpdateBooking(applicationReference, bookingRequestDto)
+    return processBookingEvents(booking, bookingRequestDto)
+  }
 
-    // application validity checks
-    applicationValidationService.validateApplication(bookingRequestDto, application, existingBooking)
+  fun updateBookedVisit(applicationReference: String, bookingRequestDto: BookingRequestDto): VisitDto {
+    val alreadyBookedVisit = visitStoreService.checkBookingAlreadyMade(applicationReference)
```
andymarke, line 68: is this logic correct given this is an update? won't this always return`alreadyBookedVisit`   for an update?

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
adaviesMOJ: As discussed via slack, leaving this as is. All logic for this refactor will remain, can look to optimise and improve in the future if needed.

Reason for the check, is a pre-existing issue where 2 API calls were made at same time for the same application, it caused issues with the 1st call creating a booking, then the 2nd call overwriting the application for that booking causing issues. The check is a crude way to control this as the Db has no locking mechanism for multi requests at same time.

PR #805, opened 2024-08-14
Db Script - Fix session slots with incorrect start and end times

adaviesMOJ:

After a script was run to move session slots to another session_template_reference we also needed to move the start and end times of the session slots.

Adding script which was run in prod to do this.


hutcheonb-moj (approved):
LGTM (as long as the query is finding the sessions we're after)

PR #807, opened 2024-08-15
Bug - Flaky test fix

adaviesMOJ:

Add a new lib to allow us to retry a known flaky test 3 times before failing.

This test fails as there is minimal concurrency contorl in place for a visit being booked at exactly the same time.


psoleckimoj (approved):
LGTM!

PR #814, opened 2024-08-23
[BUG] - Get notification groups for PVI coming back with NPE

adaviesMOJ:

When a visit is flagged, if it is a migrated visit, it will sometimes cause an error in production.

This is because the call to get notification groups has a result set of null during one of its calls to the database (which isn't allowed). This is due to the visit being a migrated visit and not having a BOOKED_VISIT event in the event audit table. It also hasn't been updated so it doesn't have an UPDATED_VISIT either.

Adding MIGRATED_VISIT to the query to make sure it never comes back null.


dhirajshettymoj (approved):
LGTM

PR #816, opened 2024-08-23
VB-4387 - Change how we process migration requests for visits in future

adaviesMOJ:

Instead of returning a 400 bad request, which forces the consumer to retry unlimited times, i'm returning a 422 specifically for this exception type, to allow the consumer to gracefully handle it.

## JIRA Link
https://dsdmoj.atlassian.net/browse/VB-4387


dhirajshettymoj (dismissed):
LGTM

dhirajshettymoj (approved):
LGTM

PR #822, opened 2024-08-28
Replace query to use join with event_audit instead of created_by column.

dhirajshettymoj:

Updates query to use join with event_audit instead of created_by column.

Implement a TODO item


adaviesMOJ (approved):
LGTM

PR #823, opened 2024-08-29
VB-3739 created by field for blocking date

dhirajshettymoj:

Add actioned_by to prison_exclude_date table

VB-3739

### DIFF
--- a/src/main/resources/db/migration/V3_28__add_actioned_by_prison_exclude_date.sql
+++ b/src/main/resources/db/migration/V3_28__add_actioned_by_prison_exclude_date.sql
@@ -1,3 +1,3 @@
 ALTER TABLE prison_exclude_date ADD actioned_by VARCHAR(60);
-update prison_exclude_date set actioned_by = 'NOT_KNOWN';
+UPDATE prison_exclude_date set actioned_by = 'NOT_KNOWN' where actioned_by IS NULL;
 ALTER TABLE prison_exclude_date ALTER COLUMN actioned_by SET NOT NULL;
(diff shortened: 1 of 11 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/integration/prison/PrisonConfigTest.kt
```diff
@@ -301,190 +282,7 @@ class PrisonConfigTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `when add exclude date called with non existent date then exclude date is successfully added`() {
```
dhirajshettymoj, line 304: moved these tests to PrisonExcludeDatesTest

src/main/resources/db/migration/V3_28__add_actioned_by_prison_exclude_date.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE prison_exclude_date ADD actioned_by VARCHAR(60);
```
adaviesMOJ, line 1: Is 60 enough, whats the limit on usernames?

adaviesMOJ (approved):
Awesome 👍

src/main/resources/db/migration/V3_28__add_actioned_by_prison_exclude_date.sql
dhirajshettymoj: usernames are usually 8. kept it 60 as username on actioned_by table also has user_name as 60.

PR #824, opened 2024-08-30
When frequency of session template is more than 4 weeks it was causing a bug of not displaying the session template

adaviesMOJ:
This was due to how we aligned the date with the closest day of the week matching the template and the booking window start date.

Rework the logic to first align the date and then loop through using the frequency to get first future date.

Currently this is a crude fix, we're looking into optimising as templates can be created many years ago and this wouldn't be suitable for those due to the while loop constantly checking.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,3 +6,4 @@
 #
 # To stop the dps-gradle-spring-boot project from overwriting any project specific customisations here, remove the
 # warning at the top of this file.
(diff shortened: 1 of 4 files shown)


dhirajshettymoj (dismissed):
LGTM

dhirajshettymoj (approved):
LGTM

tpmcgowan (approved):

dhirajshettymoj (approved):
LGTM

PR #825, opened 2024-08-30
Bug - Frequency of session templates is causing issues with displaying them

adaviesMOJ:

Alter code to first align to day of week, then loop through to past the booking window minimum date to get first bookable date


dhirajshettymoj (approved):
LGTM

PR #843, opened 2024-09-23
VB-3894 - Visitor approved and unapproved events are processed

adaviesMOJ:

The orcehstration now listens for these events.

- If unapproved event is received - flag all visits that involve that visitor and prisoner.

- If approved event is received - un-flag any previously flagged events.

## JIRA Link
https://dsdmoj.atlassian.net/browse/VB-3894


dhirajshettymoj (approved):
LGTM

PR #846, opened 2024-09-25
VB-5421 - Display the correct name for "Needs Review" page

adaviesMOJ:

Instead of returning a string to the orchestration, return all possible options and let the orchestration choose what to do for each (staff, public, system)

## JIRA Link
https://dsdmoj.atlassian.net/browse/VB-4521


dhirajshettymoj (approved):
LGTM

PR #855, opened 2024-10-01
VB-4544 - Fix Norwich issues

simonnebMOJ:
Fixes NWI new sessions with locations

Ensures correct prisons are on correct sessions


dhirajshettymoj (approved):
LGTM

PR #856, opened 2024-10-02
Create fix_LHI_sessions.sql

simonnebMOJ:

Moves sessions after the 11/10/24 to new inactive session templates with a new lower capacity.

Ensures the correct number of prisoners attend sessions


adaviesMOJ (approved):
lgtm

PR #858, opened 2024-10-07
VB-4654 - add new session_template_exclude_date table and add code to not return sessions for blocked dates.

dhirajshettymoj:

VB-4654 - adds new session_template_exclude_date table and add code to not return sessions for blocked dates.

Block individual sessions only.


adaviesMOJ (approved):
LGTM!

PR #861, opened 2024-10-09
Add booker_cancelled as new outcome status type for public bookings

adaviesMOJ:

Add booker_cancelled as new outcome status type for public bookings


dhirajshettymoj (approved):
LGTM

PR #862, opened 2024-10-10
Bug fix - Exclude reservations from the "Get single session" endpoint

adaviesMOJ:

The booked count will be used by frontend for displaying "Fully booked" session information. So we don't need to include reservations, only booked count.

## JIRA Link

https://dsdmoj.atlassian.net/browse/VB-4687


dhirajshettymoj (commented):
shall we add a test to make sure applications in progress are not counted - if it doesn't exist

hutcheonb-moj (approved):
LGTM

dhirajshettymoj (approved):
LGTM

PR #885, opened 2024-11-22
VB-4908 - Include event audit id in the HMPPS domain events

adaviesMOJ:

This id will be used by gov notify further down stream as a reference for each notification to tie it to an entry in the event_audit table.

## JIRA Ticket

https://dsdmoj.atlassian.net/browse/VB-4908


dhirajshettymoj (approved):
LGTM

PR #888, opened 2024-11-29
Remove old temporary _backup tables

adaviesMOJ:

These were added a long time ago when application tables were being introduced. They were intended to be temporary and dropped eventually. Finally cleaning these up.


dhirajshettymoj (approved):
LGTM, best run on preprod first

dhirajshettymoj (approved):
LGTM

PR #897, opened 2024-12-11
Remove duplicate call to visit-scheduler db

adaviesMOJ:

Instead of returning the visit references, just return the visits instead.


dhirajshettymoj (approved):
LGTM

PR #900, opened 2024-12-17
Gov notify callbacks - capture sent_to email or phone number

adaviesMOJ:

Add a new nullable column to the table to capture the sent_to field

Only sent via the notify callback, not the initial create


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/audit/NotifyHistoryDto.kt
```diff
@@ -19,6 +19,9 @@ data class NotifyHistoryDto(
   @Schema(description = "Notification Status", required = true)
   val status: NotifyStatus,
 
+  @Schema(description = "The email or phone number the notification was sent to", required = true)
+  val sentTo: String?,
```
dhirajshettymoj, line 23: required = false?

dhirajshettymoj (commented):
1 minor comment

dhirajshettymoj (approved):
LGTM

PR #905, opened 2025-01-02
Introduce sqs calls

dhirajshettymoj:

Add a SQSService that will allow to publish a message to the event queue and invoke it on location update so that affected visits are flagged

Add a SQSService that will allow to publish a message to the event queue and invoke it on location update so that affected visits are flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionTemplateService.kt
```diff
@@ -600,6 +613,19 @@ class SessionTemplateService(
...
+  private fun sendSessionLocationGroupUpdatedMessage(
+    prisonCode: String,
+    locationGroupReference: String,
+    oldLocations: List<PermittedSessionLocationDto>,
+  ) {
+    val sessionLocationGroupUpdatedDto = SessionLocationGroupUpdatedDto(prisonCode, locationGroupReference, oldLocations)
+    try {
+      sqsService.sendVisitEventToPrisonVisitsQueue(sessionLocationGroupUpdatedDto)
+    } catch (e: Exception) {
+      LOG.error("Error while sending SQL visit event", e)
```
adaviesMOJ, line 626: Do we need more than logging here?

adaviesMOJ (approved):
approving but worth investigating the comment about the exception handling if its needed

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/SessionTemplateService.kt
dhirajshettymoj: will leave it like this for now as the PR is being shelved.

dhirajshettymoj:
This PR will only be merged into the feature branch for future reference and not into main as this ticket has been shelved

PR #907, opened 2025-01-03
VB-4934 fix - Flag only future visits.

dhirajshettymoj:

VB-4934 fix - Replace getFutureVisitsBy with getFutureBookedVisits on VisitNotificationEventService to ensure only BOOKED visits are flagged.

VB-4934 fix

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
@@ -271,7 +271,7 @@ class VisitService(
     return eventAuditService.findByBookingReferenceOrderById(bookingReference)
   }
 
-  @Transactional
+  @Transactional(readOnly = true)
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -271,7 +271,7 @@ class VisitService(
     return eventAuditService.findByBookingReferenceOrderById(bookingReference)
   }
 
-  @Transactional
+  @Transactional(readOnly = true)
   fun getFutureVisitsBy(
```
adaviesMOJ, line 275: Is this method used anymore?

dhirajshettymoj: yes its still used on 1 enpoint /visits/search/future/{prisonerNumber} - summary - "Get future (booked and cancelled) visits for a prisoner"

PR #937, opened 2025-02-25
Fix npe visitor approved

adaviesMOJ:

Sometimes prisonCode is null, add fail safe around this issue and log when it happens


dhirajshettymoj (approved):
LGTM

PR #940, opened 2025-02-25
Change event processing to avoid NPE

adaviesMOJ:
When null comes back for some events, skip processing as it's not required. The null is because the prisonCode we have isn't active in our DB.


dhirajshettymoj (approved):
LGTM

PR #946, opened 2025-02-27
Use application-insights secret for connection string to fix deserialisation errors

adaviesMOJ:

Errors appeared after upgrading dps-spring-boot to 7.1.0. This is the fix


dhirajshettymoj (approved):
LGTM

PR #977, opened 2025-04-02
VB-5362 - add test for visit cancellations of past dates.

dhirajshettymoj:
Adds a test for past dates

Adds a test for past dates


hutcheonb-moj (approved):
lgtm

PR #978, opened 2025-04-04
HMAI-394 - New update visit from external system endpoint

tomwatkins1994:

Adds a new endpoint to allow updates to visits from external systems

This endpoint will be called by the orchestration service in the queue listener for visit writes from the HMPPS Integration API


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitService.kt
```diff
@@ -215,14 +216,34 @@ class VisitService(
...
+      updatedVisitDto.reference,
+      updatedVisitDto.createdTimestamp,
+      updatedVisitDto.modifiedTimestamp,
+      updatedVisitDto.prisonerId,
+      bookingEventAuditDto.id,
+    )
+    snsService.sendChangedVisitBookedEvent(snsDomainEventPublishDto)
+
+    return updatedVisitDto
+  }
```
dhirajshettymoj, line 237: as discussed will get back to you on these as we could possibly do away (or change) the writing of domain events  and app insights for visits that have already been booked on an external system.

tomwatkins1994: Thanks Dhiraj, I'll merge this in for now and we can always then change it later

PR #981, opened 2025-04-07
HMAI-384 - HMPPS Integration API Visit From External System Changes 

tomwatkins1994:

Merges in all of the new endpoints required for the HMPPS Integration API visits integration

Merges in all of the new endpoints required for the HMPPS Integration API visits integration


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitEventAuditService.kt
```diff
@@ -192,6 +193,10 @@ class VisitEventAuditService(private val notifyHistoryDtoBuilder: NotifyHistoryD
     bookingRequestDto: BookingRequestDto,
     eventType: EventAuditType,
   ): EventAuditDto {
+    if (bookedVisitDto.applicationReference == null) {
+      // This should not happen as application reference should always be set here
+      throw IllegalArgumentException("Visit application reference is not set")
+    }
```
dhirajshettymoj, line 199: is this part of your change or does it come from a merge?

dhirajshettymoj (commented):
1 query otherwise looks good

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/VisitController.kt
```diff
@@ -502,4 +508,139 @@ class VisitController(
...
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  fun getVisitReferenceByClientReference(
+    @Schema(description = "clientReference", example = "AABDC234", required = true)
+    @PathVariable(value = "clientReference")
+    @NotBlank
+    clientReference: String,
+  ): List<String> = visitService.getVisitReferenceByClientReference(clientReference.trim())
```
dhirajshettymoj, line 595: I think we need to restrict this to return only 1 visitReference.

tomwatkins1994: See comment on orchestration

src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitEventAuditService.kt
tomwatkins1994: We added this as we needed to satisfy the type checker in this function after we made `applicationReference` nullable. The thinking is that this exception should never be actually thrown but in the chance that somehow it ends up in this state it should be easy to trace the issue

PR #989, opened 2025-04-14
Handle issues with updating visits from external systems due to uniqueness violations

tomwatkins1994:

This should fix an issue I spotted when trying to update visits from an external system. This is the error in the log:

```
ERROR: duplicate key value violates unique constraint "visit_notes_visit_id_type_key" Detail: Key (visit_id, type)=(1013, VISITOR_CONCERN) already exists.
```
I now force a `saveAndFlush` after clearing the visit notes and before adding the new visit notes that should mean that the table gets cleared before adding new ones, thus not triggering the unique contraint.

To fix issues with updating visits from external systems


dhirajshettymoj (approved):
LGTM

PR #1012, opened 2025-05-14
Vb 5421 add exclude logic to category group

dhirajshettymoj:

Allows excluding category and incentive groups similar to location groups.

VB-5421


adaviesMOJ (approved):
LGTM!

PR #1053, opened 2025-06-26
VB-5667 Add validations exclude update session template endpoint

dhirajshettymoj:

VB-5667 - adds validation to update endpoint when include changes to exclude and vice-versa and allows users to ignore validation errors.

VB-5667 - adds validation to update endpoint when include changes to exclude and vice-versa and allows users to ignore validation errors.

### DIFF
--- a/src/main/resources/db/migration/V4_0__add_visit_sub_status_to_visit_table.sql
+++ b/src/main/resources/db/migration/V4_0__add_visit_sub_status_to_visit_table.sql
@@ -3,20 +3,21 @@
 
 -- Add sub_status column
 -- set DEFAULT to AUTO_APPROVED - this would ensure that we need not run a UPDATE for BOOKED visits as it is taking a long time on PROD
+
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/controller/admin/SessionTemplateAdminController.kt
```diff
@@ -215,7 +215,13 @@ class SessionTemplateAdminController(
     reference: String,
     @RequestBody @Valid
     updateSessionTemplateDto: UpdateSessionTemplateDto,
-  ): SessionTemplateDto = sessionTemplateService.updateSessionTemplate(reference, updateSessionTemplateDto)
+    @RequestParam(value = "validateRequest", required = false)
+    @Parameter(
```
adaviesMOJ, line 219: Why do we need to optionally validate? Were there scenarios where we didn't want to validate?

adaviesMOJ (approved):

PR #1055, opened 2025-06-27
Update visit_sub_status script as the fix needs to be applied manually due to performance issues.

dhirajshettymoj:

Temporarily update visit_sub_status script as the fix needs to be applied manually due to performance issues.

Add visit_sub_status to visit table


hutcheonb-moj (approved):
lgtm

PR #1061, opened 2025-07-15
Request visit - add new event type

adaviesMOJ:

When a visit is a request, instead of logging a VISIT_BOOKED event audit, log a REQUESTED_VISIT event audit.


tpmcgowan (approved):
LGTM

PR #1077, opened 2025-07-21
VB-5792 - Introduce app insights tracking for requested visits

adaviesMOJ:

- When a visit is requested it logs in app insights as visit-requested
- When a visit request is approved it logs in app insights as visit-request-approved


dhirajshettymoj (commented):
LGTM - but do we need to add a test for the VISIT_REQUEST_APPROVED_EVENT similar to VISIT_REQUESTED_EVENT?

dhirajshettymoj (approved):
LGTM

PR #1084, opened 2025-07-22
Change visit request approved event name to better reflect it's purpose

adaviesMOJ:

Change name to be more descriptive, as it's only raised when a visit request is approved.


dhirajshettymoj (approved):
LGTM

PR #1089, opened 2025-07-28
Stafford fixes

simonnebMOJ:

Moves visits in Stafford from multiple sessions into a single daily session

Enforcing new visit sessions timings and simplifying sessions setup


adaviesMOJ:
Not checking this in, ended up fixing via a different manual approach. Documented outside of PR.

PR #1117, opened 2025-08-22
Unflag visits which were previously flagged by appointment created event

adaviesMOJ:

This will be used by the orchestration after listening to deleted / cancelled appointment events, passing in the appointmentInstanceId, which the visit-scheduler uses to find flagged visits and un-flag them.


tpmcgowan (approved):
LGTM

PR #1119, opened 2025-08-22
Listen to appointment updated events in visits

adaviesMOJ:

visit-scheduler will use this event to un-flag existing flagged visits from the original create appointment event, and then re-search for visits to flag with the updated appointment details.


tpmcgowan (approved):
LGTM

PR #1121, opened 2025-08-28
Improve issue with endpoint eager loading collections for no reason

adaviesMOJ:

Some of the collections within event audit and other sub visit entities were being eager loaded but not used by the frontend. Create a custom DTO which only captures the required fields and stop loading of lazy collections


dhirajshettymoj (approved):
LGTM - do we need to keep it in a feature branch till the front end changes are done?

PR #1123, opened 2025-09-03
VB-5437 - Prisoner released auto rejects visit requests

adaviesMOJ:

When a prisoner released event is consumed, booked visits will be flagged and request visits will be auto rejected


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/service/VisitRequestsService.kt
```diff
@@ -91,13 +93,26 @@ class VisitRequestsService(
...
+  }
+
+  fun handlePrisonerReleasedEventAutoRejectRequestVisits(notificationDto: PrisonerReleasedNotificationDto): Int {
+    LOG.info("Entered VisitRequestsService - handlePrisonerReleasedEventAutoRejectRequestVisits")
 
+    val requestVisitsDueForAutoRejection = visitRepository.findAllVisitRequestsForPrisoner(notificationDto.prisonerNumber)
+    return processAutoRejectRequestVisits(requestVisitsDueForAutoRejection, rejectionText = "Auto rejected by prisoner released event")
+  }
+
+  private fun processAutoRejectRequestVisits(requestVisitsDueForAutoRejection: List<Visit>, rejectionText: String): Int {
```
dhirajshettymoj, line 106: suggestion only - do we want rejectionText to come from an enum? This type could also be sent to AppInsights if needed for tracking why visits have been auto rejected.

dhirajshettymoj (commented):
1 minor comment, can be ignored.

dhirajshettymoj (approved):
LGTM

PR #1126, opened 2025-09-05
VB-5437 - Request visits are auto rejected when prisoner transferred

adaviesMOJ:

Reject all request visits for prisoner outside of their current prison

## JIRA Link
https://dsdmoj.atlassian.net/browse/VB-5437


dhirajshettymoj (dismissed):
LGTM

PR #1128, opened 2025-09-09
VB-5940 - auto reject visits only when the released reason type is RELEASED

dhirajshettymoj:

VB-5940 - auto reject visits only when the released reason type is RELEASED

Fix for VB-5940 where requested visits are auto rejected when prisoner released to COURT


hutcheonb-moj (dismissed):
lgtm

hutcheonb-moj (approved):
lgtm, good test changes

PR #1132, opened 2025-09-15
VB-5924 - add validation to check if the session date is blocked when booking a visit.

dhirajshettymoj:

Introduces validation to check if the session date is blocked when booking a visit.

VB-5924


adaviesMOJ (approved):
Nice!

PR #1210, opened 2026-03-26
Upgrade deps - and fix broken integration tests

adaviesMOJ:
_Required._

Due to an increased restriction in hibernate behaviour from a dependency upgrade, we needed to move some field access inside a transaction, to avoid a 'no session' hibernate error.

## JIRA Link
https://dsdmoj.atlassian.net/browse/VB-6573


dhirajshettymoj (approved):
LGTM

PR #1228, opened 2026-05-07
VB-6650 handle add prisoner alert event

dhirajshettymoj:

VB-6650 changes to handle add prisoner alert event

Subscribe to alert events from Alerts API


adaviesMOJ (approved):
LGTM

PR #1231, opened 2026-05-11
VB-6654 handle delete prisoner alert event

dhirajshettymoj:

Handle a prisoner alert removed event

VB-6654


src/main/kotlin/uk/gov/justice/digital/hmpps/visitscheduler/dto/visitnotification/PrisonerAlertNotificationDto.kt
```diff
@@ -2,7 +2,7 @@ package uk.gov.justice.digital.hmpps.visitscheduler.dto.visitnotification
 
 import jakarta.validation.constraints.NotBlank
 
-data class PrisonerAlertUpsertedNotificationDto(
+data class PrisonerAlertNotificationDto(
```
dhirajshettymoj, line 5: renamed DTO as same DTO is used on all alert events.

adaviesMOJ (approved):

PR #1267, opened 2026-06-24
Add role to endpoint for orchestration

adaviesMOJ:

As per other endpoints, the orchestration now needs to call this endpoint - adding role to it.


tpmcgowan (approved):
LGTM

PR #1269, opened 2026-06-25
VB-6283 - Visit request rejection reason

adaviesMOJ:

Add a rejection reason to the event audit when a visit is manually rejected. The event audit free text field will be populated with the rejectionReason enum value. The frontend will map this to a set of paragraphs which can be displayed in the visit history timeline to give extra information to staff about why the visit was rejected.

## JIRA Link
https://dsdmoj.atlassian.net/browse/VB-6283


tpmcgowan (approved):
LGTM

PR #1282, opened 2026-07-10
Add validation to check if session date is valid for session template being excluded.

dhirajshettymoj:

Add validation to check if session date is valid for session template being excluded.

Update validations.


adaviesMOJ (approved):
Nice!

